PR #536, opened 2025-08-31
Harden config version validation

stranske:
## Summary
- validate config `version` string to be non-empty even without Pydantic
- record full `uv pip compile` command in requirements.lock

## Testing
- `./scripts/run_tests.sh`

------
https://chatgpt.com/codex/tasks/task_e_68b3e770d414833185f438efa2852bb2


stranske:
@ copilot, please review the issue below
The failing job is due to a SyntaxError at line 40 in src/trend_portfolio_app/sim_runner.py: “expected 'except' or 'finally' block”. This is because of an incorrectly placed logger.warning and assignment inside a try block in the compute_score_frame_local function. Here’s how to fix it:

Fix for src/trend_portfolio_app/sim_runner.py (ref: d35e39deb3a50eee2c90b32f0b16a3682a035190):

Replace lines 32–39 in compute_score_frame_local with this:

Python
for name, spec in AVAILABLE_METRICS.items():
    try:
        if name in ("sharpe", "sortino"):
            val = spec["fn"](r, idx, rf_annual)
        else:
            val = spec["fn"](r, idx)
        col_metrics[name] = val
    except Exception as e:
        logger.warning("Failed to compute metric '%s' for column '%s': %s", name, col, e)
        col_metrics[name] = np.nan
This ensures that logger.warning and assignment of np.nan are executed only if an exception is raised, resolving the syntax error.

Summary:

Move logger.warning and col_metrics[name] = np.nan into an except block.
This will allow your tests to collect and run properly.

PR #805, opened 2025-09-06
fix: assign Copilot on issue open

stranske:


stranske:

[.github/workflows/assign-to-copilot.yml](https://github.com/stranske/Trend_Model_Project/pull/805/files/3dca7b046f422b729177b5bdd58e21a3f50bdcd6#diff-20b880423d0696011a5bea0320ff40fc1b20f24903d21ece079a1c46e7cc197a)
Comment on lines +9 to +11
    if: >
      github.event.action == 'opened' ||
      contains(join(fromJson(toJson(github.event.issue.labels)).*.name, ','), 'agent:copilot')
@[chatgpt-codex-connector](https://github.com/apps/chatgpt-codex-connector) chatgpt-codex-connector bot [2 minutes ago](https://github.com/stranske/Trend_Model_Project/pull/805#discussion_r2326582699)
[P1] Guard assignment behind label even on issue open

The new job condition github.event.action == 'opened' || … causes this workflow to run for every newly opened issue, regardless of whether the issue carries the agent:copilot label. Since the script unconditionally assigns the copilot user and posts a comment when the job runs, this change will now tag Copilot on all issues, not just those explicitly labeled for Copilot. The original logic only triggered when the label was present; to support issues created with the label, the open branch should still check for that label rather than bypassing the check entirely.

PR #813, opened 2025-09-06
fix: align health check port and improve coverage

stranske:
## Summary
- target health wrapper on port 8000 in docker smoke test
- add tests for compute_score_frame_local and equity curve error handling

## Testing
- `./scripts/run_tests.sh`

------
https://chatgpt.com/codex/tasks/task_e_68bc90adb8308331a65844838fbbfce1

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -63,13 +63,13 @@ jobs:
         run: docker build -t trend:ci .
       - name: Run container
         run: |
-          docker run -d --name trendci -p 8501:8501 \
+          docker run -d --name trendci -p 8000:8000 \
             -e STREAMLIT_SERVER_HEADLESS=1 \
             trend:ci
       - name: Wait for health
         run: |
           for i in {1..10}; do
-            if curl -fsS http://localhost:8501/health | grep -qi "ok"; then
+            if curl -fsS http://localhost:8000/health | grep -qi "ok"; then
               echo "App healthy"; exit 0
             fi
             sleep 1
--- a/tests/test_sim_runner_cov.py
+++ b/tests/test_sim_runner_cov.py
@@ -156,6 +156,7 @@ def test_simulator_handles_equity_curve_update_failure(monkeypatch, caplog):
         return {"hire": [], "fire": []}
 
     def fake_apply(prev_weights, target_weights, date, rb_cfg, rb_state, policy):
+        # Intentionally assign invalid data to simulate equity curve update failure
         rb_state["equity_curve"] = 1
         return pd.Series(dtype=float)
 
--- a/tests/test_sim_runner_cov.py
+++ b/tests/test_sim_runner_cov.py
@@ -1,3 +1,4 @@
+import logging
 import numpy as np
 import pandas as pd
 import pytest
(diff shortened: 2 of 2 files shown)


stranske:

The failing job (ID: 49760165181) is caused by an assertion error in the test: tests/test_sim_runner_cov.py::test_simulator_handles_equity_curve_update_failure. The test expects the log message "Failed to update equity curve" to be present, but caplog.text is empty, causing the assertion to fail.

Solution:

Ensure that the code path actually logs "Failed to update equity curve" when the simulated equity curve update failure occurs. Check the relevant function(s) that handle equity curve updates—verify they use the logger to emit this message at the appropriate log level.
Check the test setup to confirm caplog is correctly configured to capture logs at the required level (usually ERROR or WARNING).
If the failure is simulated but no exception or log is raised, update the simulation logic so it triggers the appropriate logging.
Example Code Fix: If your update function looks like this:

Python
def update_equity_curve(...):
    try:
        # update logic
    except Exception as e:
        # Possibly missing logging
        pass
Modify to:

Python
import logging

def update_equity_curve(...):
    try:
        # update logic
    except Exception as e:
        logging.error("Failed to update equity curve")
        # other error handling
And ensure your test sets caplog to capture ERROR level:

Python
def test_simulator_handles_equity_curve_update_failure(caplog):
    caplog.set_level("ERROR")
    # rest of test...
Summary: Add or verify the error log statement in the code where the equity curve update fails, and confirm your test captures logs at the correct level. This will resolve the failed assertion. If you need to see the exact code, let me know which file or function handles "equity curve" updates, and I can give more targeted suggestions.

The failing job encountered two main issues:

1. Coverage Failure
The test coverage is 69%, which is below the required threshold of 80%.
Coverage report shows missing lines in:
src/trend_portfolio_app/sim_runner.py, specifically lines 291-292.
Solution:
Add or improve tests that exercise the code at lines 291-292 in src/trend_portfolio_app/sim_runner.py.
Ensure new or existing tests hit these lines. You can find the file here: [sim_runner.py @ c90be57a](https://github.com/stranske/Trend_Model_Project/blob/c90be57ae43c8a81b9a62ca375c60812ea6bbc4f/src/trend_portfolio_app/sim_runner.py).
2. Test Failure: Logging Assertion
The test test_simulator_handles_equity_curve_update_failure in tests/test_sim_runner_cov.py failed:
The assertion assert 'Failed to update equity curve' in caplog.text did not pass because nothing was logged.
Solution:
Confirm that the code path intended to trigger the logging of "Failed to update equity curve" is being executed.

If the code should log this message, ensure the logging statement is present and correctly configured.

Example (add to the relevant location in your code if missing):

Python
import logging
logger = logging.getLogger(__name__)
logger.error("Failed to update equity curve")
Also, verify that the test is simulating the failure correctly so the log message is emitted.

You can find the test here: [test_sim_runner_cov.py @ c90be57a](https://github.com/stranske/Trend_Model_Project/blob/c90be57ae43c8a81b9a62ca375c60812ea6bbc4f/tests/test_sim_runner_cov.py).

Summary of actions:

Increase coverage by writing tests for uncovered lines in sim_runner.py.
Fix the logging in the code and/or the test so that "Failed to update equity curve" is logged as expected.

PR #820, opened 2025-09-06
ci: set HEALTH_PORT for docker smoke

stranske:


stranske:

[.github/workflows/ci.yml](https://github.com/stranske/Trend_Model_Project/pull/820/files/faf0710065fff3d450dc05c9b0b7b1a1b2708fea#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03f)
Comment on lines 66 to +68
          docker run -d --name trendci -p 8501:8501 \
            -e STREAMLIT_SERVER_HEADLESS=1 \
            -e HEALTH_PORT=8501 \
@[chatgpt-codex-connector](https://github.com/apps/chatgpt-codex-connector) chatgpt-codex-connector bot [1 minute ago](https://github.com/stranske/Trend_Model_Project/pull/820#discussion_r2328359251)
[P1] Avoid breaking the image’s built‑in healthcheck

The new HEALTH_PORT=8501 makes the health wrapper listen on 8501, but the Docker image’s HEALTHCHECK remains hardcoded to curl http://localhost:8000/health (see Dockerfile). When this workflow runs, Docker will continually mark the container as unhealthy even though the external curl succeeds, because the internal health probe can no longer reach the service. Any step that inspects container health (e.g. docker ps --filter health=healthy or future CI gate checks) will now fail or flake. Either keep the wrapper on port 8000 and map that port, or override the container’s health command to use the same port you set here.

stranske:

Specific test failed:
tests/test_vol_drawdown_strategies.py
Test: TestStrategyIntegration.test_vol_target_factory_creation
Error:
Python
assert False
# where False = isinstance(<trend_analysis.rebalancing_file.VolTargetRebalanceStrategy object>, VolTargetRebalanceStrategy)
This means the object being tested is not recognized as an instance of VolTargetRebalanceStrategy as expected.
Solution Steps
1. Fix the Failed Test
Review the implementation and import of VolTargetRebalanceStrategy in both the test and the application code. The typical cause is either:

The object created is not of the expected class.
The import path or module is incorrect, leading to mismatched class objects.
Suggested Fix:

Ensure both the test and the code under test are importing VolTargetRebalanceStrategy from the same module.
If you have multiple definitions or shadowing in trend_analysis/rebalancing_file.py, refactor to use a single source.
Example test correction:

Python
from trend_analysis.rebalancing_file import VolTargetRebalanceStrategy

def test_vol_target_factory_creation(self):
    strategy = create_strategy()  # however your factory works
    assert isinstance(strategy, VolTargetRebalanceStrategy)
Make sure create_strategy() (or equivalent) returns the correct type.

PR #825, opened 2025-09-06
Revert "fix: unify rebalancing strategy definitions"

stranske:
Reverts stranske/Trend_Model_Project#823

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -107,7 +107,7 @@ jobs:
       - name: Wait for health
         run: |
           for i in {1..10}; do
-            if curl -fsS http://localhost:8000/health | grep -qi "ok"; then
+            if curl -fsS http://localhost:8501/health | grep -qi "ok"; then
               echo "App healthy"; exit 0
             fi
             sleep 1
--- a/src/trend_analysis/rebalancing.py
+++ b/src/trend_analysis/rebalancing.py
@@ -8,13 +8,13 @@ duplicate plugins.
 
 from __future__ import annotations
 
-from .strategies import (DrawdownGuardStrategy, DriftBandStrategy,
-                         PeriodicRebalanceStrategy,
-                         RebalancingStrategy, TurnoverCapStrategy,
-                         VolTargetRebalanceStrategy,
-                         apply_rebalancing_strategies,
-                         create_rebalancing_strategy,
-                         rebalancer_registry)
+from .rebalancing.strategies import (DrawdownGuardStrategy, DriftBandStrategy,
+                                     PeriodicRebalanceStrategy,
+                                     RebalancingStrategy, TurnoverCapStrategy,
+                                     VolTargetRebalanceStrategy,
+                                     apply_rebalancing_strategies,
+                                     create_rebalancing_strategy,
+                                     rebalancer_registry)
 
 __all__ = [
     "RebalancingStrategy",
--- a/src/trend_analysis/rebalancing.py
+++ b/src/trend_analysis/rebalancing.py
@@ -8,13 +8,13 @@ duplicate plugins.
 
 from __future__ import annotations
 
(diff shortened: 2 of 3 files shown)


stranske:
@codex please address the issues below
The job failed due to the presence of a definition for test_compute_score_frame_local_skips_date_column, which the logs indicate should be removed. To resolve this:

Locate and remove the test_compute_score_frame_local_skips_date_column definition from your codebase.
Commit and push the change to branch revert-823-codex/fix-failed-test-for-voltargetrebalancestrategy (ref: 89524879041a750c17ca473a50ae2278e84c21a8).
This should resolve the error and allow the job to complete successfully.

The failing job (ID: 49764694664) encountered this error:

Code
Remove definition: `test_compute_score_frame_local_skips_date_column`
Found 1 error.
Process completed with exit code 1.
Solution:

The test function test_compute_score_frame_local_skips_date_column either needs to be removed from your test suite or properly defined if it is missing.
If this test is obsolete or not needed, delete its definition from your codebase.
If you still need the test, ensure it is implemented correctly with the required logic.
Code suggestion (to remove the test): Find and delete the function definition, typically in your test files (e.g., test_score_frame.py):

Python
# Example: Remove or comment out this function
def test_compute_score_frame_local_skips_date_column():
    # test logic (remove this block)
    pass
Code suggestion (if you want to fix and keep the test): Make sure the function is correctly implemented with relevant assertions.

The job failed due to an ImportError while collecting tests in tests/test_rebalancing_strategies.py. The error message points to this line:

Python
fields_map = getattr(cls, "model_fields", getattr(cls, "__fields__", {}))
This likely means the test or its dependencies are trying to access model fields on a class that does not have the expected attributes ("model_fields" or "fields"). Common causes:

The class being tested may not inherit from the expected base (e.g., a pydantic Model).
There may have been a library upgrade (e.g., pydantic) that changed attribute names or inheritance patterns.
Solution Steps:

Check Class Definition: Ensure that the target class in test_rebalancing_strategies.py is inheriting from the correct base (often BaseModel for pydantic).

Example fix:

Python
from pydantic import BaseModel

class MyModel(BaseModel):
    ...
Check for Library Updates: If you recently updated pydantic or related dependencies, verify that your code is compatible with the new version. For pydantic v2, some attributes and patterns have changed.

Explicit Attribute Handling: If backward compatibility is needed, update the test code to handle both possible attribute names:

Python
if hasattr(cls, "model_fields"):
    fields_map = cls.model_fields
elif hasattr(cls, "__fields__"):
    fields_map = cls.__fields__
else:
    fields_map = {}
Next Steps:

Review your model class and the imports in tests/test_rebalancing_strategies.py.
Confirm compatibility with your pydantic (or model library) version.
Adjust the test or model code as shown above.
If you share the contents of tests/test_rebalancing_strategies.py, I can provide a more targeted code suggestion.

The job failed because the container did not respond to health checks at http://localhost:8000/health, but the logs show the app started Uvicorn on http://0.0.0.0:8501/ instead. The port mapping in the workflow is:

YAML
docker run -d --name trendci -p 8501:8000 ...
This maps container port 8000 to host port 8501, but the app inside the container is listening on 8501, not 8000. The health check is also pointed at port 8000.

Solution:

Change the health check to target port 8501:
bash
curl -fsS http://localhost:8501/health
Or, change the container to run Uvicorn on port 8000 by adjusting the app's startup command.
Recommended fix in your workflow (.github/workflows/ci.yml, ref: 89524879041a750c17ca473a50ae2278e84c21a8):

Update the health check step:

YAML
for i in {1..10}; do
  if curl -fsS http://localhost:8501/health | grep -qi "ok"; then
    echo "App healthy"; exit 0
  fi
  sleep 1
done
echo "Health check failed"; docker logs trendci; exit 1
Or ensure the app runs on port 8000 to match the health check and port mapping.

PR #979, opened 2025-09-13
ci: minimal assign-to-agent workflow fix

stranske:
Simplify workflow after SyntaxError crash.\n\nChanges:\n- Replace corrupted mixed content with clean minimal YAML.\n- Single assignment job (issues only) detects agent:codex.\n- Summary comment upsert kept minimal.\n- Separate codex_bootstrap job runs only when needed.\n- Force allow_fallback=true until PAT validated, then we can tighten.\n\nFollow-up after merge: create a labeled issue to confirm bootstrap runs to completion (no SyntaxError).


stranske:
@codex please fix comments

PR #1011, opened 2025-09-16
Add rebalancer coverage and sanitize blank ranks

stranske:
## Summary
- expand Rebalancer coverage with targeted unit tests for entry, exit, and weighting paths
- normalise blank or whitespace-only fund names before ranking to avoid empty selections
- ensure empty multi-period demo exports include a placeholder row and refresh the lockfile pins

## Testing
- ./scripts/run_tests.sh

------
https://chatgpt.com/codex/tasks/task_e_68c8c29578f08331969f58373bf6896a

### DIFF
--- a/requirements.lock
+++ b/requirements.lock
@@ -20,7 +20,7 @@ executing==2.2.1
     # via stack-data
 fonttools==4.59.2
     # via matplotlib
-hypothesis==6.138.16
+hypothesis==6.138.17
     # via trend-model (pyproject.toml)
 ipython==9.5.0
     # via ipywidgets
@@ -107,5 +107,5 @@ wcwidth==0.2.13
     # via prompt-toolkit
 widgetsnbextension==4.0.14
     # via ipywidgets
-xlsxwriter==3.2.8
+xlsxwriter==3.2.9
     # via trend-model (pyproject.toml)
--- a/scripts/run_multi_demo.py
+++ b/scripts/run_multi_demo.py
@@ -17,6 +17,7 @@ from __future__ import annotations
 
 import copy
 import importlib
+import json
 import os
 import shutil
 import subprocess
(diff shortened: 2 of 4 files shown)


stranske:
The failing job encountered this error:

Code
TypeError: Cannot read properties of undefined (reading 'listFiles')
This means your code is trying to access listFiles on something that's currently undefined. To fix it:

Solution Steps
Find the Reference to listFiles:

Identify where listFiles is called in your workflow or code.
Check the Variable/Object:

Ensure the object you're calling listFiles on is properly initialized and not undefined.
For example, if you have something like:
js
myObj.listFiles()
Make sure myObj is assigned before this line.
Add Defensive Code:

Guard the usage with a check:
js
if (myObj && typeof myObj.listFiles === 'function') {
  myObj.listFiles();
} else {
  throw new Error('myObj or myObj.listFiles is not defined');
}
Check Workflow Inputs:

If this is in a GitHub Action or workflow, verify all inputs/outputs are passed correctly to the step that uses listFiles.
Review Recent Changes:

Look at the PR or recent commits for changes that could result in myObj being undefined.
Code Suggestion
Add a safety check before calling listFiles:

js
if (yourObject && typeof yourObject.listFiles === 'function') {
  yourObject.listFiles();
} else {
  console.error('yourObject or listFiles is undefined');
}
Next Steps
Locate the exact file and line where listFiles is called (search for listFiles in your repo).
Apply the fix above and ensure the object is properly initialized.
Rerun your workflow to confirm the error is resolved.

stranske:
@codex please address the issue below
The failure in job 50449946945 is due to the following error:

could not determine current branch: failed to run git: not on any branch

This typically occurs when a workflow step tries to determine the current branch in a detached HEAD state, which is common during pull request workflows.

Solution
Update Git Command to Use Environment Variables Instead of relying on git to determine the branch, use the environment variable GITHUB_REF provided by GitHub Actions. For pull requests, this variable points to the pull request ref.

Example:

YAML
- name: Determine branch
  run: echo "Branch ref: $GITHUB_REF"
Avoid Branch-Specific Logic in Pull Request Context If your workflow or scripts require the branch name, extract it safely:

bash
BRANCH_NAME="${GITHUB_REF#refs/heads/}"
echo "Current branch: $BRANCH_NAME"
Fix in Workflow or Script Wherever you have a step or script that runs git rev-parse --abbrev-ref HEAD or similar, replace it with the above method.

Example Fix for Bash Script in Workflow:
Replace:

bash
CURRENT_BRANCH=$(git rev-parse --abbrev-ref HEAD)
With:

bash
CURRENT_BRANCH="${GITHUB_REF#refs/heads/}"
Next Steps
Update the workflow or any referenced script to use GITHUB_REF instead of git for branch determination.
Commit and push the change, then re-run the workflow.

PR #1016, opened 2025-09-16
ci(actions): switch bridge to codex-bootstrap-lite

stranske:
Update only the Codex Issue Bridge workflow to use the small local composite at .github/actions/codex-bootstrap-lite.\nNo other logic changes; same fallback path remains.\n\nMerge after PR #1015 so the action path exists on default.


stranske:
@codex fix comments

PR #1116, opened 2025-09-17
ci(autoapprove): fix pulls.listFiles undefined error

stranske:
Fix auto-approve job failure by using github.rest.pulls.listFiles and adding guards for non-PR events and missing API.\n\n- Replace github.pulls.listFiles with github.rest.pulls.listFiles (supported in actions/github-script@v7)\n- Guard when event has no pull_request payload\n- Guard when Octokit API is unavailable\n\nThis prevents TypeError: Cannot read properties of undefined (reading 'listFiles') and keeps the size/pattern gate behavior unchanged.

### DIFF
--- a/.github/workflows/autoapprove.yml
+++ b/.github/workflows/autoapprove.yml
@@ -29,7 +29,11 @@ jobs:
             }
             const pull_number = context.payload.pull_request.number;
             // Get list of files changed in the PR
-            const listFiles = github?.rest?.pulls?.listFiles || github?.pulls?.listFiles;
+            // Octokit v17+ uses github.rest.pulls.listFiles; older versions use github.pulls.listFiles.
+            // Prefer github.rest.pulls.listFiles, fallback to github.pulls.listFiles for compatibility.
+            const listFiles = github?.rest?.pulls?.listFiles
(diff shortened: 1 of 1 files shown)


stranske:
@codex Please work on the issue below. Using the variable value would be best

The failing job is caused by the test TestDemoGoldenMaster::test_coverage_gate_enforcement, which asserts that the CI workflow definition includes the string '--cov-fail-under=80'. The workflow at ref d59f34300e5f47cc16aff3a6a6b5fbf382e59d64 currently uses the variable substitution --cov-fail-under=${{ vars.COV_MIN || 80 }} instead.

Solution:
Update either the test or the workflow. The recommended fix is to adjust the test to accept the variable usage, or update the workflow to hardcode the coverage gate:

Option 1: Update the workflow
Replace line 35 in [.github/workflows/ci.yml](https://github.com/stranske/Trend_Model_Project/blob/d59f34300e5f47cc16aff3a6a6b5fbf382e59d64/.github/workflows/ci.yml) with:

YAML
35|             --cov=src --cov-report=term-missing --cov-fail-under=80 --cov-branch
O

PR #1120, opened 2025-09-17
test(ci): accept variable-based coverage threshold in CI workflow

stranske:
Relax brittle test to accept variable-based --cov-fail-under, keeping default at 80%.

### DIFF
--- a/tests/golden/test_demo.py
+++ b/tests/golden/test_demo.py
@@ -396,8 +396,11 @@ class TestDemoGoldenMaster:
         # Accept either a literal --cov-fail-under=80 or a variable-based expression resolving to 80,
         # such as --cov-fail-under=${{ vars.COV_MIN || 80 }}.
         literal_ok = "--cov-fail-under=80" in ci_content
-        variable_ok = (
-            "--cov-fail-under=${{ vars.COV_MIN" in ci_content and "|| 80" in ci_content
+        variable_ok = bool(
+            re.search(
+                r"--cov-fail-under=\$\{\{\s*vars\.COV_MIN\s*\|\|\s*80\s*\}\}",
+                ci_content,
+            )
         )
         assert (
             literal_ok or variable_ok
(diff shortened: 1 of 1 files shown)


stranske:
@codex Please confirm that the issue below has been addressed

[tests/golden/test_demo.py](https://github.com/stranske/Trend_Model_Project/pull/1120/files/be89cc7cae933e1a12216645ad387a3325d86568#diff-9f3845aad21008323f162af9de69f8dffcf7beabed747d024b8307a837bab6d2)
Outdated
Comment on lines 395 to 404
        # Verify global coverage gate is set to 80% (literal or via variable)
        # Accept either a literal --cov-fail-under=80 or a variable-based expression resolving to 80,
        # such as --cov-fail-under=${{ vars.COV_MIN || 80 }}.
        literal_ok = "--cov-fail-under=80" in ci_content
        variable_ok = (
            "--cov-fail-under=${{ vars.COV_MIN" in ci_content and "|| 80" in ci_content
        )
        assert (
            "--cov-fail-under=80" in ci_content
        ), "CI should require 80% coverage globally (found in workflow)"
            literal_ok or variable_ok
        ), "CI should require 80% coverage globally (literal or via vars.COV_MIN with default 80)"
@[chatgpt-codex-connector](https://github.com/apps/chatgpt-codex-connector) chatgpt-codex-connector bot [1 minute ago](https://github.com/stranske/Trend_Model_Project/pull/1120#discussion_r2356224444)
[P1] Ensure CI coverage gate cannot be lowered via vars.COV_MIN

The new logic now treats the presence of --cov-fail-under=${{ vars.COV_MIN || 80 }} as equivalent to a literal --cov-fail-under=80, but it never inspects the value assigned to vars.COV_MIN in the workflow. If someone sets vars.COV_MIN: 70 (or any value <80) earlier in the YAML, the pytest step will enforce only 70% coverage while this test still passes because both substrings --cov-fail-under=${{ vars.COV_MIN and || 80 remain in the file. The original test guaranteed that CI could not drop below 80%; after this change, the gate can be weakened simply by modifying the variable definition. Consider parsing the workflow to assert that any vars.COV_MIN default or assignment is ≥80 before accepting the variable-based expression.

Issue #1156, opened 2025-09-18
Perf: cache scalar per-period metrics (general metric memoization)


stranske:
### Context\nIssue #728 delivered covariance caching and incremental updates. Non-covariance scalar metrics (AnnualReturn, Sharpe, Sortino, etc.) are recomputed each period or when re-running selectors on the same window.\n\n### Goal\nIntroduce a general metric cache keyed by: (start_month, end_month, universe_hash, metric_name, config_hash) to avoid recomputation on identical windows.\n\n### Tasks\n- Implement get_or_compute_metric_series helper wrapping existing registry metrics.\n- Derive config_hash from metric parameters + risk-free + periods_per_year.\n- Add performance.cache_metrics toggle (default: false).\n- Unit tests: cache hit parity and cache disabled path.\n- Benchmark: demonstrate multi-run selector speedup.\n\n### Acceptance Criteria\n- Cache hits return identical Series (1e-12 tolerance).\n- No change to default behavior when toggle off.\n- Metrics with side effects (none currently) can be excluded gracefully.\n\n### Out of Scope\nCross-process persistence / invalidation after data mutation.\n\n### Follow-Up\nIntegrate with selector reuse (separate issue).

Issue #728, opened 2025-09-04
Performance: cache per‑period metrics and avoid redundant covariances


stranske:
Milestone: M1 Usability & Distribution
Context: Multi‑period runs recompute common pieces. Cache metrics/covs by window; vectorize turnover/cost loops.
Tasks

Memoize metrics by (start,end,universe,params) key.

Reuse covariances if only selector changes.

Replace Python loops with vectorized ops where feasible.
Acceptance criteria

Demo run wall‑clock time drops meaningfully without changing results.
Code pointers

src/trend_analysis/metrics/*, src/trend_analysis/engine/*.
Test notes

Benchmark test with a time budget; confirm output parity.

PR #1164, opened 2025-09-18
Codex bootstrap for #1156

stranske:
### Source Issue #1156: Perf: cache scalar per-period metrics (general metric memoization)

Source: https://github.com/stranske/Trend_Model_Project/issues/1156

> ### Context\nIssue #728 delivered covariance caching and incremental updates. Non-covariance scalar metrics (AnnualReturn, Sharpe, Sortino, etc.) are recomputed each period or when re-running selectors on the same window.\n\n### Goal\nIntroduce a general metric cache keyed by: (start_month, end_month, universe_hash, metric_name, config_hash) to avoid recomputation on identical windows.\n\n### Tasks\n- Implement get_or_compute_metric_series helper wrapping existing registry metrics.\n- Derive config_hash from metric parameters + risk-free + periods_per_year.\n- Add performance.cache_metrics toggle (default: false).\n- Unit tests: cache hit parity and cache disabled path.\n- Benchmark: demonstrate multi-run selector speedup.\n\n### Acceptance Criteria\n- Cache hits return identical Series (1e-12 tolerance).\n- No change to default behavior when toggle off.\n- Metrics with side effects (none currently) can be excluded gracefully.\n\n### Out of Scope\nCross-process persistence / invalidation after data mutation.\n\n### Follow-Up\nIntegrate with selector reuse (separate issue).

—
PR created automatically to engage Codex.

### DIFF
--- a/tests/test_multi_period_engine.py
+++ b/tests/test_multi_period_engine.py
@@ -450,6 +450,203 @@ def test_run_schedule_calls_rebalance_strategies_and_updates(monkeypatch):
     assert days_between > 0
 
 
+def test_threshold_hold_replacement_flow(monkeypatch: pytest.MonkeyPatch) -> None:
+    """Exercise the threshold-hold branch covering low-weight replacements."""
+
+    cfg_data = yaml.safe_load(Path("config/defaults.yml").read_text())
+    cfg_data["multi_period"] = {
+        "frequency": "M",
+        "in_sample_len": 2,
+        "out_sample_len": 1,
+        "start": "2020-01",
+        "end": "2020-04",
+    }
+
+    portfolio = cfg_data.setdefault("portfolio", {})
+    portfolio["policy"] = "threshold_hold"
+    portfolio["transaction_cost_bps"] = 25
+    portfolio["max_turnover"] = 0.1
+    th_cfg = portfolio.setdefault("threshold_hold", {})
+    th_cfg.update(
+        {
+            "target_n": 3,
+            "metric": "Sharpe",
+            "soft_strikes": 1,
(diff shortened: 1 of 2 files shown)


stranske:
@ codex please fix issues below
The failing job (ID: 50661066178) in the "autofix trivial issues" workflow is caused by an error in the script scripts/auto_type_hygiene.py. The error log points to a line involving weighting.update(scores, days=30), indicating unresolved type issues remain after an attempt to auto-fix import stubs.

Solution:

Make sure that all required type stubs are installed for your dependencies. The workflow attempts to install missing stubs using mypy --install-types --non-interactive, but some may still be missing or require manual installation.

For the specific error involving weighting.update(scores, days=30), review the type annotations for weighting and scores in your source code. Ensure that weighting has an update method accepting the scores object and a days argument of type int.

If the import for weighting or its dependencies lacks type stubs, add the module name (e.g., "weighting") to the AUTO_TYPE_ALLOWLIST environment variable so that scripts/auto_type_hygiene.py can automatically inject # type: ignore[import-untyped] into the import line. Example:

Python
import weighting  # type: ignore[import-untyped]
Or, set AUTO_TYPE_ALLOWLIST="yaml,weighting" in your environment.

If the type error is semantic (e.g., wrong argument types), update the function definition for update to match the usage:

Python
def update(self, scores: ScoreType, days: int) -> None:
    ...
Replace ScoreType with the correct type for scores.

You can view and modify scripts/auto_type_hygiene.py here:
https://github.com/stranske/Trend_Model_Project/blob/2052c87a86e2b6983cfc85c82a68465ff902ed47/scripts/auto_type_hygiene.py

PR #1269, opened 2025-09-19
Adjust agent label workflow sparse checkout guard

stranske:
## Summary
- use git ls-tree to count sparse checkout contents and only consider tracked files
- tighten the expected file count to the single trusted agent label rules file
- emit the tracked file list when the guard fails to aid future debugging

## Testing
- not run (workflow change)

------
https://chatgpt.com/codex/tasks/task_e_68ccdbaa1d508331948cd7fe4896eee1

### DIFF
--- a/.github/workflows/label-agent-prs.yml
+++ b/.github/workflows/label-agent-prs.yml
@@ -75,9 +75,12 @@ jobs:
       - name: Assert sparse checkout scope (Issue #1140)
         run: |
           set -euo pipefail
-          COUNT=$(find trusted-config -type f | wc -l | tr -d ' ')
-          if [ "$COUNT" -gt 2 ]; then
-            echo "::error::Unexpected file expansion in sparse checkout (found $COUNT files)"; exit 1;
+          EXPECTED=1
+          COUNT=$(git -C trusted-config ls-tree --full-tree --name-only HEAD | wc -l | tr -d ' ')
+          if [ "$COUNT" -ne "$EXPECTED" ]; then
+            echo "::error::Unexpected tracked file count in sparse checkout (expected $EXPECTED, found $COUNT)";
+            git -C trusted-config ls-tree --full-tree --name-only HEAD || true
+            exit 1
           fi
       - uses: actions/github-script@v7
         with:
--- a/.github/workflows/label-agent-prs.yml
+++ b/.github/workflows/label-agent-prs.yml
@@ -75,10 +75,11 @@ jobs:
       - name: Assert sparse checkout scope (Issue #1140)
         run: |
           set -euo pipefail
-          EXPECTED=1
+          # Only the agent label rules file should be present after sparse checkout.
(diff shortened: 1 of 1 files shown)


stranske:
@codex fix comments

stranske:
@codex fix the issue below
The failure is due to the use of an undefined property in your workflow file:

Code
property "suppress_activate" is not defined in object type {}
SUPPRESS_ACTIVATE: ${{ github.event.inputs.suppress_activate }}
in .github/workflows/verify-codex-bootstrap-matrix.yml line 131.

Solution
1. Define the Input in Your Workflow Make sure the workflow defines inputs.suppress_activate in the workflow_dispatch or the triggering event. For example:

YAML
on:
  workflow_dispatch:
    inputs:
      suppress_activate:
        description: 'Suppress activation'
        required: false
        default: 'false'
This ensures github.event.inputs.suppress_activate is always available.

2. Add a Default to the Expression If you want to handle missing input gracefully:

YAML
SUPPRESS_ACTIVATE: ${{ github.event.inputs.suppress_activate || 'false' }}
3. Remove or Fix the Reference If you don't need this property, remove or correct the reference:

YAML
# Remove or replace the line if the input isn't needed
Code Suggestion
Update your workflow trigger section. Example for .github/workflows/verify-codex-bootstrap-matrix.yml ([ref: f073efd3a2fde8a9a20b99a2fdbb68273826d5d7]):

YAML
on:
  workflow_dispatch:
    inputs:
      suppress_activate:
        description: 'Suppress activation'
        required: false
        default: 'false'
And use this in your job:

YAML
SUPPRESS_ACTIVATE: ${{ github.event.inputs.suppress_activate || 'false' }}
Summary:
Add the suppress_activate input definition to your workflow's dispatch section, or handle its absence with a default value. This will resolve the property-not-defined error.

stranske:
@codex address the issue below
The failing job log shows this error:

Remove repeated key literal "out_user_stats"

This means a dictionary in your code has the same key ("out_user_stats") defined more than once. Python dictionaries can't have duplicate keys—only the last occurrence will be kept, and static analysis tools (like mypy or autofix scripts) will fail if duplicates exist.

Solution:

Find the dictionary with the key "out_user_stats" repeated.
Remove or rename the duplicate key so each key is unique.
Code Suggestion:
If you have something like:

Python
my_dict = {
    "out_user_stats": value1,
    "out_user_stats": value2,  # <-- Duplicate!
}
Change it to:

Python
my_dict = {
    "out_user_stats": value2  # Keep only one definition
}
Or, if both keys are needed, use different names:

Python
my_dict = {
    "out_user_stats": value1,
    "out_user_stats_v2": value2
}
Next Steps:

Search for "out_user_stats" in your codebase, especially in files referenced by the job or workflow.
Fix any dictionary that repeats this key.
Commit and push your changes, then re-run the workflow.

PR #1277, opened 2025-09-19
Update labeler configuration to use changed-files syntax

stranske:
## Summary
- update .github/labeler.yml to use actions/labeler v5 changed-files syntax for all labels

## Testing
- ACT_DEFAULT_IMAGE=medium act pull_request -j label -W .github/workflows/pr-path-labeler.yml *(fails: docker daemon not available in environment)*

------
https://chatgpt.com/codex/tasks/task_e_68cceb5e6dec83319817e034c7e5ce4b

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,46 +1,58 @@
 # Path-based label mappings (Issue #1180)
 # Syncs labels to reflect areas of change.
 backend:
-  - 'src/**'
-  - 'trend_portfolio_app/**'
-  - 'app/**'
+  - changed-files:
+      - any-glob-to-any-file: 'src/**'
+      - any-glob-to-any-file: 'trend_portfolio_app/**'
+      - any-glob-to-any-file: 'app/**'
 
 metrics:
-  - 'src/trend_analysis/metrics.py'
-  - 'src/trend_analysis/multi_period/**'
+  - changed-files:
+      - any-glob-to-any-file: 'src/trend_analysis/metrics.py'
+      - any-glob-to-any-file: 'src/trend_analysis/multi_period/**'
 
 selection:
-  - 'src/trend_analysis/core/**'
-  - 'src/trend_analysis/engine/**'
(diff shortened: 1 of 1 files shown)


stranske:
@codex please address the cause of the failing tests below
The failing job (50755608986) in your CI workflow is caused by multiple test failures, with the key error being:

Code
ValueError: not enough values to unpack (expected 2, got 0)
Affected tests:

test_merge_update_recurses_through_nested_dicts
test_summarise_multi_handles_missing_fields
test_to_yaml_preserves_key_order
test_summarise_run_df_rounds_numeric_columns
test_read_defaults_sets_csv_path
These all fail with the same unpacking error, indicating that a function is trying to unpack a sequence (likely a tuple or list) of length 0, but expects 2 values.

Solution Steps
Locate the Problematic Function These tests are likely using a helper function that unpacks values from an iterable. Look for code like:

Python
a, b = some_iterable
or

Python
for k, v in some_dict.items():
Add Defensive Checks Ensure that the iterable has the expected number of items before unpacking. Example fix:

Python
if len(some_iterable) == 2:
    a, b = some_iterable
else:
    # handle empty or malformed case
Or catch the error and handle gracefully:

Python
try:
    a, b = some_iterable
except ValueError:
    # handle when iterable is empty
    pass
Update the Helper Function If the problematic code is inside a helper (for example, a dict merging or summarising function), update it to handle cases where input might be empty.

Example Code Fix
Suppose in src/trend_portfolio_app/io_utils.py, you have:

Python
def merge_update(d1, d2):
    for k, v in d2.items():
        # recursive logic
Change to:

Python
def merge_update(d1, d2):
    for item in d2.items():
        if len(item) == 2:
            k, v = item
            # recursive logic
        else:
            continue  # skip if not a key/value pair
Or, for tuple unpacking:

Python
try:
    k, v = item
except ValueError:
    continue
Next Steps
Identify and update the relevant helper function(s) used by the failing tests.

PR #1315, opened 2025-09-19
Fix workflow inputs scope

stranske:
## Summary
- nest workflow_dispatch inputs under the correct on.workflow_dispatch trigger in verify-codex-bootstrap-matrix
- keep schedule and push triggers unchanged while ensuring suppress_activate and other inputs resolve correctly

## Testing
- not run

------
https://chatgpt.com/codex/tasks/task_e_68cdb42ab9488331a0cb6bb74941904f

### DIFF
--- a/.github/workflows/verify-codex-bootstrap-matrix.yml
+++ b/.github/workflows/verify-codex-bootstrap-matrix.yml
@@ -2,13 +2,6 @@ name: Verify Codex Bootstrap Matrix
 
 on:
   workflow_dispatch:
-  schedule:
-    - cron: '17 3 * * *'
-  push:
-    paths:
-      - '.github/workflows/verify-codex-bootstrap-matrix.yml'
-      - 'scripts/verify_codex_bootstrap.py'
-      - 'docs/codex-simulation.md'
     inputs:
       scenarios:
         description: "Comma-separated list: all or subset. See scripts/verify_codex_bootstrap.py SCENARIOS_IMPL for available scenarios."
@@ -38,6 +31,13 @@ on:
         description: "Set to true for t13_suppressed"
         required: false
         default: "false"
+  schedule:
+    - cron: '17 3 * * *'
+  push:
+    paths:
+      - '.github/workflows/verify-codex-bootstrap-matrix.yml'
+      - 'scripts/verify_codex_bootstrap.py'
+      - 'docs/codex-simulation.md'
 
 jobs:
   plan:


stranske:
@codex fix error below

The job failed because the workflow used the unsupported replace() function in this line:

YAML
scenario: ${{ fromJson(format('["{0}"]', replace(needs.plan.outputs.scenario_list, ',', '","'))) }}
GitHub Actions expressions do not support replace(). To fix this, use only supported functions. A common approach is to move string manipulation into a shell script step, or split using available features.

Solution:
Replace the matrix definition in your workflow as follows:

YAML
strategy:
  fail-fast: false
  matrix:
    scenario: ${{ fromJson('["' + join(needs.plan.outputs.scenario_list, '","') + '"]') }}
Or, if scenario_list is already a comma-separated string, convert it in a step before the matrix is used, or output it as a JSON array from your compute step.

Summary: Remove replace() and use supported functions (like join()), or format scenario_list as a JSON array in your output.
See the failing line in your workflow here:
[.github/workflows/verify-codex-bootstrap-matrix.yml#L70 (ref: 813bcf25a53821971731254bd196b69497fbcba9)](https://github.com/stranske/Trend_Model_Project/blob/813bcf25a53821971731254bd196b69497fbcba9/.github/workflows/verify-codex-bootstrap-matrix.yml#L70)

stranske:
@codex fix type errors below
The failing job (ID: 50817281469) in your autofix workflow encountered Python type errors when running mypy type checks. The log states:

Found 4 errors (2 fixed, 2 remaining).
The failure is related to type annotations, specifically with TypeAlias annotations resolving to list[float].
Exit code 1 indicates mypy found blocking type issues.
Solution Steps:

Locate Remaining Type Errors:

Open the file(s) referenced in the error message (look for lines around comment:
Code
# ``TypeAlias`` annotations resolve to ``list[float]`` at runtime on
at line 40).
Fix TypeAlias Definitions:

Ensure your TypeAlias definitions use the correct syntax for both type checking and runtime. Example for Python 3.10+:
Python
from typing import TypeAlias

FloatList: TypeAlias = "list[float]"
For older Python versions, use:
Python
from typing import List

FloatList = List[float]
Run mypy Locally:

After fixing, run mypy locally to confirm all type errors are resolved:
Code
mypy .
Commit and Push Fixes:

Commit your changes and push to the PR branch.
Code Suggestion Example: If the error is at line 40 of a file, and the TypeAlias is incorrectly defined, update it as follows:

Before:

Python
# TypeAlias annotations resolve to list[float] at runtime
MyListType: TypeAlias = list[float]
After:

Python
from typing import TypeAlias

MyListType: TypeAlias = "list[float]"  # For Python 3.10+
Or, if you support older Python:

Python
from typing import List

MyListType = List[float]
Summary:

Fix TypeAlias syntax in the referenced file(s).
Ensure compatibility with your Python version.
Re-run mypy and commit the fixes.

PR #1316, opened 2025-09-19
chore(ci): enforce coverage gate via pytest flag

stranske:
## Summary
- ensure the reusable Python CI workflow fails immediately when coverage drops below the configured gate by passing `--cov-fail-under`
- retain the existing coverage summary step so diagnostics still emit the canonical gate message

## Testing
- pytest tests/golden/test_demo.py::TestDemoGoldenMaster::test_coverage_gate_enforcement -k coverage --maxfail=1

------
https://chatgpt.com/codex/tasks/task_e_68cdbaf0493c8331b8b2196a405c95b3

### DIFF
--- a/.github/workflows/reusable-ci-python.yml
+++ b/.github/workflows/reusable-ci-python.yml
@@ -117,9 +117,9 @@ jobs:
           set -o pipefail
           MARK='${{ steps.markers.outputs.value }}'
           if [ "${{ inputs.run-full }}" = "true" ] || [ -z "$MARK" ]; then
-            pytest --junitxml=pytest-junit.xml --cov=src --cov-report=term-missing --cov-report=xml:coverage.xml --cov-branch 2>&1 | tee pytest.log || echo $? > test_status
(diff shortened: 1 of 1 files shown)


stranske:
@codex fix comments
The job failed because the GitHub Actions workflow at .github/workflows/verify-codex-bootstrap-matrix.yml attempts to use the function replace in an expression:

YAML
scenario: ${{ fromJson(format('[\"{0}\"]', replace(needs.plan.outputs.scenario_list, ',', '\",\"'))) }}
However, replace is not a supported function in GitHub Actions expressions. Supported functions include always, cancelled, contains, endswith, failure, format, fromjson, hashfiles, join, startswith, and success.

Solution:
To split a comma-separated list into a JSON array, use the join function instead of replace. Here’s how you can fix the expression:

YAML
scenario: ${{ fromJson(format('[\"{0}\"]', join('\",\"', needs.plan.outputs.scenario_list))) }}
This will join the items in needs.plan.outputs.scenario_list using "","", producing a valid JSON array string for fromJson.

Summary of actions:

Remove the use of replace.
Use the join function as shown above.
Update line 70 in the workflow file to use the corrected expression. This will resolve the job failure.

PR #1318, opened 2025-09-19
Add tests for multi-period schedule update and proxy error handling

stranske:
## Summary
- extend multi-period engine helper tests to cover Series rebalance inputs and the schedule update path
- add proxy server tests for uvicorn startup, websocket failures, and byte query handling

## Testing
- pytest tests/test_multi_period_engine_helpers_additional.py -q
- pytest tests/test_proxy_server_additional.py -q

------
https://chatgpt.com/codex/tasks/task_e_68cdc3ac76ac8331b858e2216509cb1e

### DIFF
--- a/tests/test_multi_period_engine_helpers_additional.py
+++ b/tests/test_multi_period_engine_helpers_additional.py
@@ -93,5 +93,83 @@ def test_portfolio_rebalance_accepts_multiple_input_shapes() -> None:
     assert set(pf.history[feb_key].index) == {"F1", "F3"}
     assert pf.costs[feb_key] == pytest.approx(0.02)
 
+    # Direct Series input should be preserved without copying columns.
+    march_series = pd.Series({"F2": 0.7, "F4": 0.3}, dtype=float)
+    pf.rebalance("2021-03-31", march_series, turnover=0.12)
+
+    march_key = "2021-03-31"
+    assert pf.history[march_key].equals(march_series.astype(float))
+    assert pf.turnover[march_key] == pytest.approx(0.12)
+
     # Costs accumulate over time.
     assert pf.total_rebalance_costs == pytest.approx(0.07)
+
+
+def test_run_schedule_invokes_update_and_fast_turnover(monkeypatch: pytest.MonkeyPatch) -> None:
+    """``run_schedule`` should exercise fast turnover and update hooks."""
+
+    class DummySelector:
+        rank_column = "Sharpe"
+
+        def select(self, score_frame: pd.DataFrame) -> tuple[pd.DataFrame, pd.DataFrame]:
+            return score_frame, score_frame
+
+    class RecordingWeighting:
+        def __init__(self) -> None:
+            self.updates: list[tuple[pd.Series, int]] = []
+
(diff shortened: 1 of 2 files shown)


stranske:
@codex please address the type error below
The job failed due to unresolved type errors; specifically, there were remaining issues after running mypy and the auto_type_hygiene.py script.

Solution:

Review the files reported in the logs for unresolved type errors. The log mentions "Found 4 errors (2 fixed, 2 remaining)"—these are likely type issues not covered by the script.
For imports of untyped modules (such as yaml), ensure the line includes the comment # type: ignore[import-untyped]. The script auto_type_hygiene.py is designed to inject this comment for allowed modules, but if your errors are for imports not in the allowlist, add them to the AUTO_TYPE_ALLOWLIST environment variable or manually append the ignore comment.
If the errors are not import-related, examine the specific type errors mypy reports and resolve them by:
Adding missing type annotations.
Installing any required type stubs via pip (for example: pip install types-PACKAGE).
Adjusting function signatures or variable types to match expected types.
Example code fix for import errors:

Python
import yaml  # type: ignore[import-untyped]
For further details, see the full script here: [scripts/auto_type_hygiene.py (ref: c1074bcbbaca0f5eae74a2e0a7e1dcf204da3694)](https://github.com/stranske/Trend_Model_Project/blob/c1074bcbbaca0f5eae74a2e0a7e1dcf204da3694/scripts/auto_type_hygiene.py).

Summary:

Add # type: ignore[import-untyped] to untyped imports.
Resolve remaining type errors mypy reports.
Expand AUTO_TYPE_ALLOWLIST if other modules need ignoring.
Ensure all required type stubs are installed.

stranske:
@codex please address the issue below
The failing job (50821315633) error is caused by a failed test:

Test: tests/test_proxy_server_additional.py::test_handle_http_request_accepts_non_string_query
Failure:
Python
assert False
+  where False = <built-in method endswith of str object at 0x7fa00297a790>('/metrics?id=42')
+    where <built-in method endswith of str object at 0x7fa00297a790> = "http://example.com:1234/metrics?b'id=42'".endswith
Problem Analysis
The test expects a string URL ending with /metrics?id=42, but the actual URL is "http://example.com:1234/metrics?b'id=42'", with a b prefix indicating a bytes object was converted to a string incorrectly.
This usually happens when code does str(bytes_obj) instead of bytes_obj.decode().
Solution
Check the code where the query is added to the URL.
Likely in the proxy server code (possibly in a request handler function).

Code Fix Example
Suppose you have something like:

Python
url = f"http://example.com:1234/metrics?{str(query)}"
Replace with:

Python
url = f"http://example.com:1234/metrics?{query.decode() if isinstance(query, bytes) else str(query)}"
Or, if you always expect query to be bytes:

Python
url = f"http://example.com:1234/metrics?{query.decode()}"
Test Fix (if the code is correct and test is wrong)
If your code correctly handles bytes and the test expects a string, change the test input to use a string instead of bytes when calling the handler.

Next Steps
Locate the code in your proxy server (likely in src/trend_analysis/api_server or related handler files) where the query is processed and appended to the URL.
Update the code to properly decode bytes to string before formatting the URL.

PR #1322, opened 2025-09-19
Provide yaml stub so imports remain typed without ignores

stranske:
## Summary
- add an in-repo typing stub for the yaml module so mypy recognizes safe_load/dump and related exceptions
- teach scripts/auto_type_hygiene.py to skip adding import-untyped ignores when a stub or py.typed marker is present
- remove the temporary import-untyped ignores from config, GUI, and CI probe modules now that yaml has typing support

## Testing
- mypy .
- python scripts/auto_type_hygiene.py

------
https://chatgpt.com/codex/tasks/task_e_68cdcb80fd048331a3919c32de2c7ff5

### DIFF
--- a/scripts/auto_type_hygiene.py
+++ b/scripts/auto_type_hygiene.py
@@ -4,7 +4,8 @@
 Goals (automation-safe):
   * Install missing type stubs (handled by workflow separately) – this script only mutates source files.
   * Inject `# type: ignore[import-untyped]` comments for a curated allowlist of known untyped imports
-    (currently only `yaml`).
+    (currently only `yaml`).  If a stub file or `py.typed` marker is present, the script leaves the import
+    unchanged.
   * Idempotent: never duplicate ignore comments.
   * Skip legacy / excluded paths (Old/, notebooks/old/).
   * Avoid masking *semantic* errors – we only touch import lines that succeed at runtime but lack stubs.
@@ -23,6 +24,7 @@ This script intentionally uses only the standard library.
 """
 from __future__ import annotations
 
+import importlib.util
 import os
 import re
 from pathlib import Path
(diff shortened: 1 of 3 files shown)


stranske:
@codex fix comments

stranske:
@codex please identify the cause of the failing test below and fix
The failing job is caused by the test: tests/test_proxy_server_additional.py::test_handle_http_request_accepts_non_string_query.

Failure Details:
The assertion error is:

Code
assert False
+  where False = <built-in method endswith of str object at 0x7f8c1bdf60d0>('/metrics?id=42')
+    where <built-in method endswith of str object at 0x7f8c1bdf60d0> = "http://example.com:1234/metrics?b'id=42'".endswith
This indicates that the code is generating a URL like "http://example.com:1234/metrics?b'id=42'", which is incorrect because the query parameter is being encoded as a bytes string (b'id=42') instead of a regular string.

Solution: You should ensure that query parameters in your HTTP request handler are decoded to strings before building the URL.

Code Fix Suggestion:
Find the code in your proxy server that handles incoming queries and constructs the request path.
If you see code like:

Python
query = b'id=42'
url = f"http://example.com:1234/metrics?{query}"
Change this to:

Python
if isinstance(query, bytes):
    query = query.decode()   # Ensure query is a string
url = f"http://example.com:1234/metrics?{query}"
Or, if your framework uses request objects, ensure you convert bytes to string before using or passing them to URL construction.

Where to Apply:
Apply this fix in the code that is tested by test_handle_http_request_accepts_non_string_query—likely in your proxy server implementation. You can locate the code by searching for the test name or the related request handler in your repository.

Reference:

Workflow file: [.github/workflows/ci.yml](https://github.com/stranske/Trend_Model_Project/blob/ba52916d2106fec759a08f38a335e194c99f2258/.github/workflows/ci.yml)
Test file: [tests/test_proxy_server_additional.py](https://github.com/stranske/Trend_Model_Project/blob/ba52916d2106fec759a08f38a335e194c99f2258/tests/test_proxy_server_additional.py)
Summary:
Decode any bytes query parameters to string before constructing URLs in your proxy server handler. This will fix the test and resolve the job failure. If you want exact code lines for the failing handler, let me know.

PR #1323, opened 2025-09-20
Fix proxy HTTP query handling for byte queries

stranske:
## Summary
- normalize HTTP proxy query strings by decoding byte-valued request.url.query attributes before forwarding

## Testing
- ./scripts/run_tests.sh

------
https://chatgpt.com/codex/tasks/task_e_68ce13720fbc8331b5f56d1afe22ce94

### DIFF
--- a/src/trend_analysis/proxy/server.py
+++ b/src/trend_analysis/proxy/server.py
@@ -202,8 +202,18 @@ class StreamlitProxy:
         _assert_deps()
         normalized = path if path.startswith("/") else f"/{path}"
         target_url = urljoin(self.streamlit_base_url, normalized)
-        if q := str(request.url.query):
-            target_url += f"?{q}"
+        query = getattr(request.url, "query", "")
+        if isinstance(query, bytes):
+            try:
+                decoded_query = query.decode()
+            except UnicodeDecodeError:
+                decoded_query = query.decode("latin-1", errors="ignore")
+        elif query is None:
+            decoded_query = ""
+        else:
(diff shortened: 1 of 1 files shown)


stranske:
@codex please address the type errors below
The job failed due to type errors reported in a Python script—specifically related to TypeAlias annotations resolving to list[float]. The log mentions:

“Found 4 errors (2 fixed, 2 remaining).”
The script involved is likely scripts/auto_type_hygiene.py, but it may have been skipped if missing.
mypy type stubs installation was attempted, but errors persisted.
Solution Steps:

Review TypeAlias Annotations

Check for TypeAlias usage in files that define type aliases for list[float].
Ensure all aliases are properly defined and used consistently.
Fix Remaining Type Errors

Run mypy locally to see the exact error messages and their locations.
For any TypeAlias issues, define the alias as follows (example):
Python
from typing import TypeAlias

FloatList: TypeAlias = list[float]
Replace any ambiguous or incorrect alias usages with the correct type or alias.
Ensure Type Stubs Are Installed

If using external libraries, install missing stubs:
Code
pip install types-<library>
Re-run mypy after installing any missing stubs.
Code Suggestion Example: If you have:

Python
# Incorrect usage
FloatList = list[float]
Change to:

Python
from typing import TypeAlias

FloatList: TypeAlias = list[float]
Next Steps:

Locate the file(s) with the remaining errors (search for TypeAlias and list[float]).
Apply the corrections above.
Commit and push the fixes, then re-run the workflow.

PR #1327, opened 2025-09-20
Include chatgpt parser script in Docker builds

stranske:
## Summary
- adjust the Docker ignore rules to include the ChatGPT topics parser script

## Testing
- pytest -q

------
https://chatgpt.com/codex/tasks/task_e_68ce1ee0db08833196f681c402bc9aa3

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -74,7 +74,9 @@ docker-compose.override.yml
 Dockerfile.dev
 
 # CI/CD
-.github/
+.github/*
+!.github/scripts/
+!.github/scripts/parse_chatgpt_topics.py
 
 # Local configuration
 .env.local
--- a/.dockerignore
+++ b/.dockerignore
@@ -76,8 +76,6 @@ Dockerfile.dev
 # CI/CD
 .github/*
 !.github/scripts/
-!.github/scripts/parse_chatgpt_topics.py
-
 # Local configuration
 .env.local
 .env.*.local


stranske:
@codex please address the type error mentioned below
to unresolved type errors during the execution of scripts/auto_type_hygiene.py. The logs indicate:

The script found 4 errors (2 fixed, 2 remaining).
The job ended with exit code 1 after running type hygiene and mypy stub installation steps.
Solution:

Review the specific type errors reported by mypy. Since the script only auto-fixes missing stubs for imports in the allowlist (default: yaml), ensure that all untyped imports are either in the allowlist or have available stubs.
Check that your source files do not have semantic type errors (undefined names, return mismatches, etc.), as the script does not auto-ignore these.
If the errors are due to missing type stubs for external modules, add those modules to the AUTO_TYPE_ALLOWLIST environment variable or install the appropriate stubs (e.g., via pip install types-<modulename>).
If you need to bypass unresolved import typing for other modules, modify the allowlist in the workflow environment or directly in your CI configuration.
Code Suggestions:

To expand your allowlist, set in your workflow YAML:
YAML
env:
  AUTO_TYPE_ALLOWLIST: "yaml,other_module"
To install missing stubs:
bash
pip install types-<modulename>
For further review, see the hygiene script here: [scripts/auto_type_hygiene.py](https://github.com/stranske/Trend_Model_Project/blob/5586873b7a7709ff48a61181d19bacdaa85187f0/scripts/auto_type_hygiene.py).

PR #1328, opened 2025-09-20
Docs/ci workflow guide

stranske:


stranske:
@codex please fix the error below
The job failed because the workflow attempted to use the replace() function in the matrix definition, but replace is not a valid function in GitHub Actions expressions. The error occurred at line 70 of the workflow file:

YAML
70| scenario: ${{ fromJson(format('["{0}"]', replace(needs.plan.outputs.scenario_list, ',', '","'))) }}
Solution:
Replace the unsupported replace() function with the correct syntax. You can use the split() function to convert a comma-separated string into an array, which is suitable for a matrix definition.

Code Suggestion:
Update line 70 to:

YAML
scenario: ${{ split(needs.plan.outputs.scenario_list, ',') }}
This change will properly create an array of scenarios for the matrix job.
See the workflow file here: [.github/workflows/verify-codex-bootstrap-matrix.yml@8ad25d9702f4c7dd2a76cac2770cce84bcd017cf](https://github.com/stranske/Trend_Model_Project/blob/8ad25d9702f4c7dd2a76cac2770cce84bcd017cf/.github/workflows/verify-codex-bootstrap-matrix.yml)

PR #1360, opened 2025-09-20
Add coverage soft gate automation to CI workflow

stranske:
## Summary
- update the reusable Python CI workflow to emit coverage artifacts and JUnit XML for every matrix job
- introduce a coverage soft gate job that aggregates artifacts, generates a hotspots summary, updates the coverage tracking issue, and publishes the data to the workflow summary
- ensure the overall CI gate waits for the soft gate job while keeping coverage from blocking merges

## Testing
- not run (workflow changes only)

------
https://chatgpt.com/codex/tasks/task_e_68cee549d588833183f54f4938016e1c

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,6 +31,7 @@ jobs:
       - name: Download coverage artifacts
         uses: actions/download-artifact@v4
         with:
+          run-id: ${{ needs.reuse.outputs.run_id || github.run_id }}
           pattern: coverage-*
           merge-multiple: true
           path: coverage_artifacts
@@ -93,7 +94,8 @@ jobs:
           else:
             header = "**Coverage data unavailable**\n\n"
 
-          hotspots = sorted(file_stats.items(), key=lambda item: item[1])[:15]
+          HOTSPOT_LIMIT = 15  # Capture the 15 lowest-covered files to keep the table concise.
+          hotspots = sorted(file_stats.items(), key=lambda item: item[1])[:HOTSPOT_LIMIT]
 
           with coverage_path.open('w', encoding='utf-8') as handle:
             handle.write(header)
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,9 +18,220 @@ jobs:
       # Coverage gate enforced in the reusable workflow via
       # pytest --cov --cov-fail-under=${{ vars.COV_MIN || 85 }}
       run_quarantine: 'false'
+  coverage_soft_gate:
+    name: coverage soft gate (file/update issue)
+    needs: [reuse]
+    if: ${{ always() }}
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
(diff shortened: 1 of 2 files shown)


stranske:
@codex fix comments

stranske:
@codex fix comments

stranske:
The job failed due to a JavaScript error: "SyntaxError: Identifier 'core' has already been declared." This happened in a GitHub Actions workflow script, likely within a Node.js context. The error message indicates that the variable core is declared more than once in the same scope.

How to fix it:

Review the workflow script (likely in .github/workflows/ci.yml or any referenced JavaScript/TypeScript action file) and ensure that core is only imported or declared once. For example, if you have:

js
const core = require('@actions/core');
// ... other code ...
const core = require('@actions/core'); // <-- This will trigger the error
Remove the duplicate declaration so only one import exists.

If using ES modules, check for duplicate import statements:

js
import * as core from '@actions/core';
// ... other code ...
import * as core from '@actions/core'; // <-- This will trigger the error
Only import core once at the top of the file.

Code suggestion: Remove any duplicate lines like:

js
const core = require('@actions/core');
or

js
import * as core from '@actions/core';
so each file has only a single declaration.

Next steps:

Search the workflow's JavaScript files for multiple core declarations.
Fix by keeping only one declaration per file.

stranske:
@codex please identify the causes of the failing tests mentioned below and fix them
The job failed due to two test failures in tests/test_multi_period_engine_threshold_events_extended.py:

test_threshold_hold_event_log_and_replacements
test_threshold_hold_weight_bounds_deficit
The logs show repeated errors involving:

Code
try:
    from ..perf.cache import CovCache
    cov_cache_obj = CovCache()
except Exception:
    cov_cache_obj = None
and

Code
if cov_cache_obj is not None:
    cov_cache_obj.incremental_updates += 1
These errors suggest CovCache cannot be imported or initialized correctly, likely due to an import path issue or missing module.

Solution Steps
1. Check Module Import Path
Ensure that perf/cache.py exists and that the relative import from ..perf.cache is correct. If the test is not running in a package context, consider changing the import to absolute:

Python
# Instead of:
from ..perf.cache import CovCache
# Try:
from perf.cache import CovCache
Or, adjust your test runner to execute tests in a way that supports relative imports (using python -m pytest from the repo root).

2. Verify CovCache Implementation
Confirm CovCache is defined in perf/cache.py and is not missing any dependencies.

3. Defensive Fallback
If CovCache is optional, ensure your fallback logic does not suppress real errors. Consider logging the actual exception for diagnosis:

Python
try:
    from perf.cache import CovCache
    cov_cache_obj = CovCache()
except Exception as e:
    print("CovCache import/init failed:", e)
    cov_cache_obj = None
Code Suggestion
Update your import statement and add exception logging as above. If the relative import is required, ensure your directory structure and test runner support it.

Next Steps
Fix the import in tests/test_multi_period_engine_threshold_events_extended.py.

PR #1406, opened 2025-09-21
fix: address workflow validation errors

stranske:
## Summary
- fix the auto-approve workflow step layout so the verification runs inside the step block
- emit the Codex bootstrap scenario list as JSON and reuse it for both matrix and sequential modes
- tidy reusable CI coverage tables so empty rows use explicit n/a placeholders

## Testing
- not run (workflow-only changes)

------
https://chatgpt.com/codex/tasks/task_e_68cf81d5c7708331adc7178451cb914b


stranske:
@codex please fix the error below
The failure in job 50867154744 is caused by a YAML syntax error in the workflow file .github/workflows/verify-codex-bootstrap-matrix.yml. The error message from the logs is:

could not parse as YAML: yaml: line 58: could not find expected ':' [syntax-check] 58 | import json, os

This indicates that on line 58 of the workflow file, there is a line starting with "import json, os", which is Python syntax and not valid YAML. In GitHub Actions workflow files, Python code should not be placed directly in the YAML file—it must be part of a script section, typically under a run: key in a step.

Solution:

Ensure that Python code (like import json, os) is inside a run: block within a step.
Example fix:
YAML
# Instead of this (invalid)
58 | import json, os

# Use this (valid)
steps:
  - name: Run Python script
    run: |
      import json, os
      # your python code here
Next Steps:

Edit .github/workflows/verify-codex-bootstrap-matrix.yml.
Move any Python code to a run block under a step.
Ensure YAML indentation is correct and all colons are present where expected.
After these changes, commit and push your fix to the branch associated with pull request #1406.

PR #1463, opened 2025-09-21
Add shift-safe signal helpers and causal tests

stranske:
## Summary
- add shift-safe `compute_signal` helper to the pipeline module so signals always lag inputs by one period
- introduce `position_from_signal` to translate signals into positions sequentially without look-ahead
- cover the helpers with deterministic and property-based tests that fail if future data leaks into earlier decisions

## Testing
- pytest tests/test_shift_safe_pipeline.py

------
https://chatgpt.com/codex/tasks/task_e_68d05ed5f6188331919b7722f8021b13

### DIFF
--- a/src/trend_analysis/pipeline.py
+++ b/src/trend_analysis/pipeline.py
@@ -664,6 +664,68 @@ def run_full(cfg: Config) -> dict[str, object]:
     return {} if res is None else res
 
 
+# --- Shift-safe helpers ----------------------------------------------------
+
+
+def compute_signal(
+    df: pd.DataFrame,
+    *,
+    column: str = "returns",
+    window: int = 3,
+    min_periods: int | None = None,
+) -> pd.Series:
+    """Return a trailing rolling-mean signal that is shift-safe.
+
+    The returned signal is *explicitly* shifted by one period so the value at
+    index ``t`` only depends on data strictly before ``t``.  This avoids the
+    classic off-by-one bug where the latest observation leaks into the
+    decision made at the same timestamp, which would introduce look-ahead
+    bias in downstream portfolio simulations.
+    """
+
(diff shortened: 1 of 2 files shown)


stranske:
@codex fix comments

stranske:
@codex please identify the cause of the failing tests below and fix
The failing job (#50888780515) is caused by the following error in multiple tests:

Code
This error occurs when code in src/trend_analysis/config/models.py tries to import with:

Python
from .model import validate_trend_config
But the module tests.model doesn't exist or isn't discoverable.

Solution
1. Check the Import Path

If the intent is to import from src/trend_analysis/config/model.py, update the import to:
Python
from .model import validate_trend_config
Ensure that model.py is present in src/trend_analysis/config/.

If the import in your tests is from .model import ... but the test file is not inside a package, Python may resolve it incorrectly. Consider using an absolute import.
2. Fix Relative Import in Tests

In tests/test_config_fallback_additional.py (and any other affected test), change:

Python
from .model import validate_trend_config
To:

Python
from src.trend_analysis.config.model import validate_trend_config
Or, if you want to keep relative imports, ensure your test directory has an __init__.py file so it's treated as a package.

3. Add Missing File if Needed

If model.py is missing in tests/, but all logic is in src/trend_analysis/config/model.py, remove the incorrect import from your tests and use the correct one as above.

Example Code Fix for Tests

Python
# At the top of tests/test_config_fallback_additional.py
from src.trend_analysis.config.model import validate_trend_config
4. Ensure PYTHONPATH

Make sure your test runner sets PYTHONPATH to the project root so that src is discoverable.

Summary:
Update all imports from from .model import ... to the correct absolute path referencing src/trend_analysis/config/model.py, and ensure your test files and directories are structured as packages if using relative imports.

PR #1466, opened 2025-09-21
Enforce shift-safe signal helpers and causal validation

stranske:
## Summary
- implement shift-safe signal generation and sequential positioning helpers in the pipeline
- add deterministic and property-based tests that fail if the helpers read future information
- adjust configuration fallback loader to import validation helpers robustly and skip Pydantic validation when unavailable

## Testing
- `pytest -q`

------
https://chatgpt.com/codex/tasks/task_e_68d07f8751cc8331906382f2b3bb292a

### DIFF
--- a/src/trend_analysis/config/models.py
+++ b/src/trend_analysis/config/models.py
@@ -14,7 +14,11 @@ from typing import Any, ClassVar, Dict, List, Protocol, cast
 
 import yaml
 
-from .model import validate_trend_config
+# ``models.py`` is executed under different module names in the test suite so we
+# import ``validate_trend_config`` via its fully-qualified path to avoid
+# relative-import resolution against the temporary alias (for example when the
+# module is loaded as ``tests.config_models_fallback``).
+from trend_analysis.config.model import validate_trend_config
 
 
 class ConfigProtocol(Protocol):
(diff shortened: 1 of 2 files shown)


stranske:
@codex fix comments

stranske:
@codex identify the cause of the failing tests below and fix
he failing job is caused by the following error in your test suite:

Error:
FAILED tests/test_config_models_additional.py::test_load_config_rejects_invalid_type - UnboundLocalError: cannot access local variable 'cfg_dict' where it is not associated with a value

Root Cause:
This error means the variable cfg_dict is referenced in your test but has not been assigned before access. This typically happens if an exception or conditional prevents its assignment.

Solution:

Review the test function test_load_config_rejects_invalid_type in tests/test_config_models_additional.py.
Ensure cfg_dict is assigned in all code paths before it is used.
Example fix:
Python
def test_load_config_rejects_invalid_type():
    cfg_dict = None  # Initialize to avoid UnboundLocalError
    try:
        # Your code that may assign cfg_dict
        cfg_dict = load_config(...)
    except SomeExpectedException:
        pass
    assert cfg_dict is not None  # or adjust based on your logic
Recommended next steps:

Review and update the test as above.

stranske:
@codex Please fix the code formatting issues below
The failing job (ID: 50891821637) is caused by code formatting issues. According to the logs, two files would be reformatted, specifically including:

src/trend_analysis/config/models.py
The job is running a "Style Gate" check, which usually means tools like black, flake8, or similar are enforcing style rules.

Solution: Run the code formatter locally and commit the changes. If you are using black, you can run:

bash
black src/trend_analysis/config/models.py
Then, commit the reformatted files and push them to the branch associated with your pull request.

Summary of steps:

Run the appropriate formatter (e.g., black) on src/trend_analysis/config/models.py and any other files flagged.
Commit and push the changes.
This will resolve the style gate failure and allow the workflow to pass. If you need specific style settings, check your repo’s config files (pyproject.toml or setup.cfg).

If you want to see the changes black would make, you can run:

bash
black --diff src/trend_analysis/config/models.py
If you need more details or the exact file diff, let me know!

PR #1468, opened 2025-09-21
Fix config fallback import and extend coverage

stranske:
## Summary
- guard the configuration fallback import to avoid package resolution errors when tests simulate missing pydantic and ensure a safe shim is provided
- add regression coverage that exercises the fallback loader after the production module has already been imported so the shim is executed in CI

## Testing
- pytest --maxfail=1 --disable-warnings -q

------
https://chatgpt.com/codex/tasks/task_e_68d08222bce08331a02d326a6f773ed5

### DIFF
--- a/src/trend_analysis/config/models.py
+++ b/src/trend_analysis/config/models.py
@@ -21,22 +21,22 @@ import yaml
 # ``tests.config_models_fallback``) which breaks the relative ``.model`` import.
 # Falling back to the absolute import keeps both the production import path and
 # the isolated test harness working.
+def _fallback_validate_trend_config(_data: Mapping[str, Any], *, base_path: Path) -> None:  # type: ignore[override]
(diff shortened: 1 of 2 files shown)


stranske:
@codex run black on the files identified below
The failing job is due to style issues: 6 files would be reformatted, causing the style gate to exit with code 1. This means your code does not meet the formatting standards enforced by your workflow.

Solution:

Run the code formatter locally (likely black or a similar tool, based on the message) on your branch:
bash
black .
Commit and push the changes:
bash
git add .
git commit -m "fix: reformat code to pass style gate"
git push
Key file example:
One of the files mentioned in the logs:

tests/test_logging_module.py
Why this works:
The "would reformat" message means the formatter found code not matching the style requirements. The workflow expects all files to be formatted before merging.

PR #1473, opened 2025-09-22
Improve merge manager allowlist matching

stranske:
## Summary
- expand the merge manager glob implementation so allowlists accept **, *, and ? patterns while caching compiled regexes
- allow the workflow to auto-approve safe, low-risk PRs once provenance labels are present even before the `automerge` label is applied

## Testing
- pytest *(fails: ModuleNotFoundError: No module named 'tests.model' when running config fallback tests in the environment)*

------
https://chatgpt.com/codex/tasks/task_e_68d09222689c8331bde8b304fb197cfc

### DIFF
--- a/.github/scripts/merge_manager.js
+++ b/.github/scripts/merge_manager.js
@@ -27,14 +27,49 @@ async function fetchAllowlist(github, owner, repo, path, ref) {
   return { found, patterns, maxLines };
 }
 
+const patternCache = new Map();
+
+function escapeRegexSegment(segment) {
+  return segment.replace(/([\\^$*+?.()|{}\[\]])/g, '\\$1');
+}
+
+function compileGlob(pattern) {
+  if (patternCache.has(pattern)) {
+    return patternCache.get(pattern);
+  }
+  let regex = '';
+  for (let i = 0; i < pattern.length; i += 1) {
+    const char = pattern[i];
+    if (char === '*') {
+      if (pattern[i + 1] === '*') {
+        if (pattern[i + 2] === '/') {
+          regex += '(?:.*/)?';
+          i += 2;
+        } else {
+          regex += '.*';
+          i += 1;
+        }
+      } else {
+        regex += '[^/]*';
+      }
+      continue;
+    }
+    if (char === '?') {
+      regex += '[^/]';
+      continue;
+    }
+    regex += escapeRegexSegment(char);
+  }
+  const compiled = new RegExp(`^${regex}$`);
+  patternCache.set(pattern, compiled);
+  return compiled;
+}
+
 function matchPattern(filename, pattern) {
-  if (pattern.endsWith('/**')) {
-    return filename.startsWith(pattern.slice(0, -3));
+  if (!pattern || !filename) {
+    return false;
   }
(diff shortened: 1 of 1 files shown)


stranske:
@codex fix the errors described below
The job failed because of repeated errors in test setup:
This error occurs in tests/test_config_fallback_additional.py, where code tries to run from .model import validate_trend_config but cannot find model.py in the tests directory.

Solution:

Make sure there is a model.py file in the tests directory. If it’s missing, create it or restore it from version control.
If validate_trend_config is defined elsewhere (like in src/trend_analysis/config/model.py), update the import in tests/test_config_fallback_additional.py:
Python
# Change this:
from .model import validate_trend_config
# To this:
from src.trend_analysis.config.model import validate_trend_config
Ensure your test files use the correct import paths, especially if you are using relative imports.
Code Suggestion for tests/test_config_fallback_additional.py: Replace any line like:

Python
from .model import validate_trend_config
With:

Python
from src.trend_analysis.config.model import validate_trend_config
Or fix the imports to match the actual location of the function.

Also Check:

If your test runner is configured to include src in the Python path. If not, update your test runner or add a PYTHONPATH=src in your workflow configuration.
Reference:

[src/trend_analysis/config/model.py (ref: 8cf0a970339d32d279784010780a5f83ad234975)](https://github.com/stranske/Trend_Model_Project/blob/8cf0a970339d32d279784010780a5f83ad234975/src/trend_analysis/config/model.py)

stranske:
@codex fix the error below
The failing job is caused by multiple errors like:

Code
This happens when code in src/trend_analysis/config/models.py tries to run:

Python
from .model import validate_trend_config
but the tests.model module does not exist.

Solution:

Check that the import path for validate_trend_config is correct in src/trend_analysis/config/models.py (line 17).
If you meant to import from src/trend_analysis/config/model.py, update the import to:
Python
from .model import validate_trend_config
(if you are inside the same package).
If the code is being run from tests, and you need to import from the main package, use:
Python
from trend_analysis.config.model import validate_trend_config
assuming your package is structured so that trend_analysis is importable.
Code suggestion: Replace any incorrect relative import in src/trend_analysis/config/models.py (line 17) with the correct absolute or relative path for validate_trend_config. For example:

Python
# src/trend_analysis/config/models.py
from trend_analysis.config.model import validate_trend_config
Or, if you are inside the config package:

Python
from .model import validate_trend_config
Next steps:

Verify the existence of model.py within src/trend_analysis/config/.
Ensure your test runner sets up PYTHONPATH to include src/ so imports resolve correctly.

stranske:
@codex run black on the files indicated below
The job failed because the code style check detected files that need reformatting. Specifically, the following error was reported:

4 files would be reformatted, 350 files would be left unchanged.

Most notably, the file /tests/test_logging_module.py was flagged. The style checker (likely Black or similar) exits with a non-zero code when files are not properly formatted.

Solution:

Run your code formatter locally to auto-format the files:

bash
black .
or, to format just the flagged file:

bash
black tests/test_logging_module.py
Commit and push the reformatted files. This will allow the job to pass, as the style gate will find no formatting violations.

Example command to fix:

PR #1477, opened 2025-09-22
Add unified trend CLI entrypoint

stranske:
## Summary
- add a dedicated `trend` package with a consolidated CLI for running the pipeline, generating reports, launching the app, and executing canned stress windows
- document the new entrypoint, wire a development wrapper script, and expose the console script through `pyproject.toml`
- cover the CLI helpers with focused unit tests and keep the setup script aware of the new wrapper

## Testing
- pytest tests/test_trend_cli.py
- pytest tests/test_script_error_handling.py

------
https://chatgpt.com/codex/tasks/task_e_68d096f8e5d08331abe068dac1b3f05a

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,11 +45,13 @@ Install the latest stable release from PyPI:
 pip install trend-model
 ```
 
-This provides the ``trend-model`` command with both GUI and pipeline modes:
+This provides both the legacy ``trend-model`` command and the new unified
+``trend`` entrypoint with pipeline, reporting, stress testing, and app launch
+modes:
 
 ```bash
-trend-model run --help
-trend-model gui
+trend run --help
+trend app
 ```
 
 ### Via ``pipx``
@@ -58,7 +60,7 @@ For an isolated installation without activating a virtual environment:
 
 ```bash
 pipx install trend-model
-trend-model gui
+trend app
 ```
 
 ### From Source
@@ -78,10 +80,10 @@ If package installation fails due to network issues, the CLI is still available 
 ```bash
 # Use the development wrapper script
 ./scripts/trend-model run --help
-./scripts/trend-model gui
+./scripts/trend app
 
-# Or run the module directly  
-PYTHONPATH="./src" python -m trend_analysis.cli run --help
+# Or run the module directly
+PYTHONPATH="./src" python -m trend.cli run --help
 ```
 
 ### Docker (Zero-Setup)
@@ -93,7 +95,7 @@ For the fastest setup with zero local dependencies:
 docker run -p 8501:8501 ghcr.io/stranske/trend-model:latest
 
 # Use the CLI
-docker run --rm ghcr.io/stranske/trend-model:latest trend-analysis --help
+docker run --rm ghcr.io/stranske/trend-model:latest trend --help
 ```
 
 Then visit http://localhost:8501 for the interactive web interface.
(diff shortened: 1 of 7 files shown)


stranske:
@codex run black on the files below
The failing job is caused by style issues detected by a code formatter (likely Black or a similar tool). The logs show:

10 files would be reformatted, including tests/test_walkforward_engine.py.
The workflow exited with code 1 because files do not meet style requirements.
Solution: Run the following command locally in your repository to automatically fix style issues:

sh
black .
This will reformat all Python files to comply with the required style. After running it, commit and push the changes.

If you only want to fix the file mentioned in the logs:

sh
black tests/test_walkforward_engine.py
Then commit and push.

Summary:
Reformat your code using the project's preferred formatter (most likely Black) and push the changes to resolve the job failure.

stranske:
@codex fix the issues described below
The failing job (50894928154) is caused by mypy type-checking errors in two files:

1. src/trend_analysis/config/models.py
Lines 29 & 36:
Code
error: Name "_fallback_validate_trend_config" is not defined  [name-defined]
Solution:
Ensure that _fallback_validate_trend_config is defined or imported in this file.
Python
# At the top of src/trend_analysis/config/models.py
from .validators import _fallback_validate_trend_config  # if defined elsewhere
Or define the function if missing.
2. src/trend/cli.py
Line 22:

Code
error: Unused "type: ignore" comment  [unused-ignore]
Solution:

Remove the unnecessary # type: ignore comment on line 22.
Line 27:

Code
error: Incompatible types in assignment (expression has type "None", variable has type "Callable[[object], dict[str, int] | None]")  [assignment]
Solution:

Make sure the variable is assigned a function with the correct signature, not None.
Python
def my_func(obj: object) -> dict[str, int] | None:
    ...
variable: Callable[[object], dict[str, int] | None] = my_func
Line 29:

Code
error: All conditional function variants must have identical signatures  [misc]
The two function signatures for maybe_log_step and _legacy_maybe_log_step must match.
Solution:

Refactor so both functions have the same signature:
Python
def _legacy_maybe_log_step(enabled: bool, run_id: str, event: str, message: str, **fields: Any) -> None:
    ...
Line 125:

Code
error: Argument 1 to "load_csv" has incompatible type "Path"; expected "str"  [arg-type]
Solution:

Pass a string to load_csv:
Python
load_csv(str(path_variable))
Line 308:

Code
error: Function "_legacy_extract_cache_stats" could always be true in boolean context  [truthy-function]
Solution:

Use explicit checks:
Python
if _legacy_extract_cache_stats() is not None:
    ...
Next Steps
Fix the missing import/definition of _fallback_validate_trend_config in [src/trend_analysis/config/models.py](https://github.com/stranske/Trend_Model_Project/blob/8eb6389ac694f0d37cbdbf092151f5880df0c434/src/trend_analysis/config/models.py).
Clean up type annotations, function signatures, and argument types in [src/trend/cli.py](https://github.com/stranske/Trend_Model_Project/blob/8eb6389ac694f0d37cbdbf092151f5880df0c434/src/trend/cli.py) as described above.

stranske:
@codex please identify the cause of the failing tests below and fix
The failing job reports two main issues:

NameError: name '_fallback_validate_trend_config' is not defined

This error appears multiple times, especially in src/trend_analysis/config/models.py and the related tests in tests/test_config_fallback_additional.py.
Solution: Make sure the _fallback_validate_trend_config function is defined and imported correctly in src/trend_analysis/config/models.py. If it is meant to come from another module, confirm the import path and that the function exists.
Example fix:

Python
# src/trend_analysis/config/models.py
from .fallback import _fallback_validate_trend_config  # Adjust the import path as needed

This error is triggered in src/trend_analysis/config/models.py at line 25.
Solution: Verify if tests/model.py exists. If it’s test code, update the import to use relative import (for example, from ..tests.model import ...), or remove the dependency from production code if not needed.
Example fix:

Python
# Avoid importing test modules in production code
# src/trend_analysis/config/models.py, line 25
# Remove or replace:
# import tests.model
Assertion and Index Errors in Tests

Example: assert np.float64(nan) == 0.0 ± 1.0e-12 fails due to a NaN value.
Solution: Check the logic producing NaN; ensure input data or calculations do not result in NaN. Use numpy.isnan() to assert the value is not NaN before comparing.
Example fix:

Python
import numpy as np
assert not np.isnan(value), "Value is NaN"
assert value == pytest.approx(expected)
Summary of actions needed:

Define and correctly import _fallback_validate_trend_config.
Remove test-module imports from production code.
Fix NaN-producing test/data logic.
Ensure test dependencies exist and are properly referenced.

PR #1489, opened 2025-09-22
Fix config fallback validation integration

stranske:
## Summary
- ensure the fallback validator is used when pydantic is unavailable while still capturing stub validation calls
- wrap stub validators so tests that instrument validate_trend_config can observe the base_path without raising errors

## Testing
- pytest

------
https://chatgpt.com/codex/tasks/task_e_68d0daed54a48331a4e38e18a75559d4

### DIFF
--- a/src/trend_analysis/config/models.py
+++ b/src/trend_analysis/config/models.py
@@ -140,53 +140,8 @@ def _fallback_validate_trend_config(
 try:  # pragma: no cover - exercised indirectly via tests
     from .model import validate_trend_config
 except ImportError:  # pragma: no cover - defensive fallback for test harness
-    if sys.modules.get("pydantic") is None:
-
-def _fallback_validate_trend_config(
-    data: Mapping[str, Any] | None,
-    *,
-    base_path: Path,
-) -> None:
-    """Best-effort configuration validation used when Pydantic is unavailable.
-
-    The full :mod:`trend_analysis.config.model` helper performs rich validation
-    with Pydantic.  When that module cannot be imported (for example, during
-    unit tests that intentionally remove the dependency) we still want to catch
-    obvious schema issues so downstream code receives a sensible error instead
-    of failing later.  Only lightweight checks are performed here – the
-    fallback ``Config`` class below performs the detailed validation when it is
-    instantiated.
-    """
-
-    if not isinstance(data, Mapping):
-        raise TypeError("Configuration payload must be a mapping")
-
-    version = data.get("version")
-    if not isinstance(version, str):
-        raise ValueError("version must be a string")
-    if not version or not version.strip():
-        raise ValueError("Version field must be a non-empty string")
-
-    required_sections = [
-        "data",
-        "preprocessing",
-        "vol_adjust",
-        "sample_split",
-        "portfolio",
-        "metrics",
-        "export",
-        "run",
-    ]
-    missing = [section for section in required_sections if section not in data]
-    if missing:
-        raise ValueError(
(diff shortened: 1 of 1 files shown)


stranske:
@codex fix type error below
1 reference
The failing job encountered a mypy error in [src/trend_analysis/config/models.py](https://github.com/stranske/Trend_Model_Project/blob/95fedc2011d9c7025a98367cac97c69d4ddef2f0/src/trend_analysis/config/models.py) at line 206:

Code
src/trend_analysis/config/models.py:206: error: Name "validate_trend_config" already defined (possibly by an import)  [no-redef]
Cause:
The error occurs because the name validate_trend_config is defined multiple times in the file. Specifically, it is imported at the top (lines 141–142) and then assigned again at line 206:

Python
206| validate_trend_config: _ValidateConfigFn = _resolve_validate_trend_config()
Solution:
To fix the mypy [no-redef] error, you should rename the variable assignment at line 206 to avoid clashing with the imported name. For example, you can use a different name for the resolved function:

Python
# Replace this line:
validate_trend_config: _ValidateConfigFn = _resolve_validate_trend_config()

# With this:
resolved_validate_trend_config: _ValidateConfigFn = _resolve_validate_trend_config()
Then, update any usage of validate_trend_config in this file to use resolved_validate_trend_config instead.

Alternatively, if the intent is to override the imported name, you can suppress the mypy warning using # type: ignore[no-redef]:

Python
validate_trend_config: _ValidateConfigFn = _resolve_validate_trend_config()  # type: ignore[no-redef]
Recommendation:
Prefer renaming to avoid confusion and maintain clarity in the codebase. This will resolve the type checking error and allow the workflow to pass.

Reference:

[src/trend_analysis/config/models.py (ref: 95fedc2011d9c7025a98367cac97c69d4ddef2f0)](https://github.com/stranske/Trend_Model_Project/blob/95fedc2011d9c7025a98367cac97c69d4ddef2f0/src/trend_analysis/config/models.py)

stranske:
@codex find sources of failing tests and fix
The failing job encountered multiple test failures, mainly in these files:

tests/test_workflow_agents_consolidation.py
tests/test_workflow_codex_issue_bridge.py
tests/test_workflow_merge_manager.py
Common Failure Causes
The failures are primarily related to missing workflow steps, triggers, or workflow existence checks. These errors often result from:

Outdated or missing workflow YAML files.
Changes in workflow structure not reflected in the tests.
The code or configuration not matching test expectations.
Solution Steps
Check Workflow Definitions
Review the YAML files for the following workflows in your repo:

Agents consolidation workflow
Codex issue bridge workflow
Merge manager workflow
Ensure each workflow:

Exists at the expected location.
Contains all required triggers and steps referenced in the tests.
Synchronize Tests and Workflows If the workflows were changed or renamed, update either the YAML files or the test functions so their expectations match.
Example: If test_merge_manager_core_steps_present fails due to missing steps, verify your workflow contains those steps.

Example Fix for a Missing Step If a test expects a step named commit-prefix-quote in .github/workflows/merge-manager.yml, add it:

YAML
steps:
  - name: Commit Prefix Quote
    run: echo "Ensuring commit prefix is quoted"
Address Skipped Tests The lockfile consistency test is skipped because uv CLI is missing.
Install it in your Dockerfile or runner:

Dockerfile
RUN pip install uv
General Recommendations

Ensure Docker builds use a consistent Python environment matching your local setup.
If running as root, consider switching to a non-root user for pip installs to avoid permission warnings.
Next Steps
Review and update the workflows and their corresponding tests.
Rerun the CI to confirm the failures are resolved.

PR #1501, opened 2025-09-22
fix(agents): fallback Codex dispatch on bootstrap failure

stranske:
Adds workflow_dispatch fallback to codex-issue-bridge when primary bootstrap fails.

### DIFF
--- a/.github/workflows/assign-to-agents.yml
+++ b/.github/workflows/assign-to-agents.yml
@@ -239,5 +239,40 @@ jobs:
       - name: Surface bootstrap failure
         if: steps.ctx.outputs.should_run == 'true' && steps.ctx.outputs.target_type == 'issue' && steps.ctx.outputs.agent == 'codex' && steps.bootstrap.outcome == 'failure'
         run: |
-          echo "::error::Codex bootstrap failed."
+          echo "::warning::Codex bootstrap step failed (non-fatal while fallback evaluated)."
+
+      - name: Fallback dispatch Codex Issue Bridge
+        id: fallback_dispatch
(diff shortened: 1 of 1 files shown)


stranske:
@codex find the cause of the failing tests mentioned below and fix
The job failed due to a missing required field: "version" in the configuration passed to a Pydantic model (_PydanticConfigImpl). This error occurs in multiple tests, all reporting:

Code
pydantic_core._pydantic_core.ValidationError: 1 validation error for _PydanticConfigImpl
version
  Field required [type=missing, input_value=..., input_type=dict]
Solution:
You need to ensure that the configuration (often loaded from YAML or JSON) includes a "version" field at the top level. This applies to your config files such as config/demo.yml and any programmatically constructed config dictionaries.

Steps to fix:

Add version to your config files
For example, in config/demo.yml:

YAML
version: "1.0"
data:
  csv_path: demo/demo_returns.csv
  warmup_periods: 0
# ... other config fields
Check code that builds config dictionaries
If you build config dicts in code, add "version":

Python
config = {
    "version": "1.0",
    "data": { "csv_path": ..., "warmup_periods": ... }
    # ...
}
Update tests that use minimal or custom configs
Any test that constructs a config without "version" should be updated.

Code suggestion for config loader or test setup:
If you want to default the version inside your config loader, you could add:

Python
config.setdefault("version", "1.0")
But explicit is better than implicit—prefer adding "version" directly in config files.

Summary:
Add the required "version" field to all config files and config dicts passed to _PydanticConfigImpl. This will resolve the validation error and allow your workflow to pass.

For more details, see the Pydantic error docs: https://errors.pydantic.dev/2.11/v/missing

PR #1506, opened 2025-09-22
Fix config version validation failures and tighten autofix guard

stranske:
## Summary
- ensure GUI test fixtures write demo config files with the required `version` key
- tighten the workflow_run autofix guard so it only skips when github-actions pushes an autofix-prefixed commit
- document the standardized prefix + actor guard coupling in the workflows README

## Testing
- pytest tests/test_app_coverage.py tests/test_gui_app_extended.py *(fails: ModuleNotFoundError: No module named 'joblib')*

------
https://chatgpt.com/codex/tasks/task_e_68d1a1864fd08331a3f5b2c539e58613

### DIFF
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -157,7 +157,7 @@ Acceptance Criteria (Issue #1415) satisfied by: archival of legacy workflows, pr
 ### 7.1 Autofix Loop Guard (Issue #1347)
 Loop prevention layers:
 1. The consolidated workflow only reacts to completed CI runs (no direct `push` trigger).
-2. Guard logic inspects the latest commit subject for the configured prefix (default `chore(autofix):`) when the actor is `github-actions`.
+2. Guard logic only fires when the workflow actor is `github-actions` (or `github-actions[bot]`) **and** the latest commit subject begins with the standardized prefix `chore(autofix):`.
 3. Style Gate runs independently and does not trigger autofix.
 
 Result: Each human push generates at most one autofix patch sequence; autofix commits do not recursively spawn new runs.
--- a/.github/workflows/autofix.yml
+++ b/.github/workflows/autofix.yml
@@ -129,8 +129,11 @@ jobs:
               const subject = (commit.data.commit.message || '').split('\n')[0];
               result.head_subject = subject;
               const actor = result.actor;
-              if (subject.toLowerCase().startsWith(prefix.toLowerCase())) {
-                core.info('Loop guard engaged: detected prior autofix commit.');
+              const isAutomation = actor === 'github-actions' || actor === 'github-actions[bot]';
+              const subjectLower = subject.toLowerCase();
+              const prefixLower = prefix.toLowerCase();
+              if (isAutomation && prefixLower && subjectLower.startsWith(prefixLower)) {
+                core.info(`Loop guard engaged for actor ${actor}: detected prior autofix commit.`);
                 result.loop_skip = 'true';
               }
             } catch (error) {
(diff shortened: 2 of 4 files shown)


stranske:
@codex fix errors below
The failing job log shows several shell script and YAML syntax errors in your workflow files. Here are focused fixes for the main problems:

1. Shell Script Errors in .github/workflows/autofix.yml
Error: Couldn't find 'fi' for this 'if', Couldn't parse this if expression, Expected 'fi' matching previously mentioned 'if'
Solution: Review all if statements in your workflow scripts. Make sure every if has a matching fi, and that the syntax is correct.
bash
if [ condition ]; then
  # commands
fi
Example Fix:
bash
# Incorrect
if [ "$foo" = "bar" ]
  echo "Foo is bar"
# Missing 'then' and 'fi'

# Correct
if [ "$foo" = "bar" ]; then
  echo "Foo is bar"
fi
2. Double Quote to Prevent Globbing and Word Splitting
Error: SC2086: Double quote to prevent globbing and word splitting
Solution: Always quote shell variables:
bash
# Incorrect
rm $files

# Correct
rm "$files"
3. YAML Syntax Error in .github/workflows/verify-codex-bootstrap-matrix.yml
Error: yaml: line 58: could not find expected ':'
Solution: On line 58, the line import json, os is not valid YAML. If this is meant to be a command, it must be part of a run: block and properly formatted.
YAML
steps:
  - name: Run Python Import
    run: |
      python -c "import json, os"
4. Workflow Input Type Error in .github/workflows/ci.yml
Error: Input "enable-soft-gate" and "coverage-hard-fail" are typed as string, but a bool value was assigned.
Solution: Always pass boolean-like values as strings in workflow inputs:
YAML
enable-soft-gate: 'true'
coverage-hard-fail: 'false'
Ensure that the reusable workflow expects strings, not booleans.
5. Style Suggestion for Redirection
Error: SC2129: Consider using { cmd1; cmd2; } >> file instead of individual redirects
Solution: Where you have multiple lines redirecting output to a file, group them:
bash
{ echo "line1"; echo "line2"; } >> output.txt
Links to Failing Code
[autofix.yml (fix if statements and quoting)](https://github.com/stranske/Trend_Model_Project/blob/80ec266a3a463f1c37199c3a97c41a045090c7ac/.github/workflows/autofix.yml)
[verify-codex-bootstrap-matrix.yml (fix YAML syntax)](https://github.com/stranske/Trend_Model_Project/blob/80ec266a3a463f1c37199c3a97c41a045090c7ac/.github/workflows/verify-codex-bootstrap-matrix.yml)
[ci.yml (fix workflow input types)](https://github.com/stranske/Trend_Model_Project/blob/80ec266a3a463f1c37199c3a97c41a045090c7ac/.github/workflows/ci.yml)

stranske:
@codex fix issue below
The job failed due to style issues: 5 files need reformatting according to your style gate check. The failure message indicates that running the style checker (likely black or similar) flagged these files but did not automatically reformat them.

Solution: Run the style formatting tool locally (e.g., black . if using Black) and commit the changes. Specifically, the log mentions one file:

tests/test_multi_period_engine_price_frames_extra.py
But 5 files in total need reformatting.

Code Suggestion: If you use Black, run this in your repository root:

bash
black .
Then stage and commit the reformatted files:

bash
git add .
git commit -m "Apply code formatting to pass style gate"
git push
Summary:
Format the code in the flagged files using your repository's expected formatter, then commit and push the changes. This will resolve the style gate failure and allow the job to pass.

PR #1517, opened 2025-09-22
chore(codex): align autofix guard policy

stranske:
## Summary
- add an actor + commit prefix guard to the scheduled residual cleanup workflow so it respects the standardized autofix loop protection
- document that all autofix lanes share the same `chore(autofix):` prefix and guard behaviour

## Testing
- not run (workflows and docs change only)

------
https://chatgpt.com/codex/tasks/task_e_68d1b5dbad988331b1a614bb1dd6aa27

### DIFF
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -93,7 +93,9 @@ jobs:
       cov_min: 70
 ```
 Autofix commits use the configurable prefix (default `chore(autofix):`). The consolidated workflow guards against loops by
-detecting automation actors + existing prefix and only running after the CI workflow completes.
+detecting automation actors + existing prefix and only running after the CI workflow completes. Scheduled cleanup and reusable
+autofix helpers consume the same prefix so the guard behaviour is identical no matter which workflow authored the last automation
+commit.
 
 ```yaml
 name: Agents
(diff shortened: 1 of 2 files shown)


stranske:
@codex fix the comments below
The failure in job 50979579558 is due to incorrect input types in your workflow. The reusable workflow .github/workflows/reusable-ci-python.yml expects the inputs enable-soft-gate and coverage-hard-fail to be strings, but the values provided are boolean.

Solution: Update .github/workflows/ci.yml so that the inputs are strings (wrapped in quotes):

YAML
enable-soft-gate: "true"
coverage-hard-fail: "false"
This change ensures the values match the expected input types of the reusable workflow.

Additionally, there's a ShellCheck warning in .github/workflows/verify-ci-stack.yml at line 28: Double quote to prevent globbing and word splitting. Update your script as follows:

Example fix:

bash
run: |
  some_command "$variable"
Always quote variables in shell scripts to prevent word splitting and globbing issues.

Apply these fixes to resolve the failing workflow. If you need the exact code snippet from your workflow files, let me know!

stranske:
@codex fix the mypy error below
The failing job (50979579828) encountered a mypy error:

Error: src/trend_analysis/config/models.py:28: error: Unused "type: ignore" comment [unused-ignore]

Solution: Remove the unused "# type: ignore" comment on line 28 in src/trend_analysis/config/models.py. mypy flagged this because there is no error to ignore on that line, so the comment is unnecessary and causes the type check to fail.

Code suggestion:

Open src/trend_analysis/config/models.py at ref 515929752d774da73606983cc684f1c36d4b4f40.
Delete or revise the "# type: ignore" on line 28.
Once you remove this comment, rerun the workflow to confirm the fix. This should resolve the mypy failure and allow the job to complete successfully.

stranske:
@codex identify the cause of failing tests below and fix
The failing job 50979579822 is caused by two test failures in tests/test_config_models_fallback_loader.py:

test_fallback_loader_activates_when_imports_fail
test_fallback_loader_returns_minimal_payload
Both fail with: ValueError: version must be a string

Solution: Review the fallback loader logic in your config models code. Ensure that any payload or config dictionary always sets the "version" key as a string, not as another type (e.g., int, None, etc.).

Suggested Fix: Find where the fallback payload is created (likely in src/trend_analysis/config/models.py or a related loader function). Update it like this:

Python
payload['version'] = str(payload.get('version', '0'))
Or, when you build the minimal payload for fallback, ensure:

Python
minimal_payload = {
    'version': '0',  # Always a string!
    ...
}
Next Steps:

Audit any place where "version" is set.
Update tests to expect version as a string.
Rerun CI after making this change.

PR #1518, opened 2025-09-22
Docker smoke reliability: buildx cache summary logging

stranske:
## Summary
- rename the Docker workflow's build job to `smoke` and keep the 10 minute timeout
- key the buildx cache strictly on the runner OS plus Dockerfile and requirements.lock hashes
- write the buildx cache hit or miss status to both workflow notices and the job summary

## Testing
- ./actionlint -color .github/workflows/docker.yml

------
https://chatgpt.com/codex/tasks/task_e_68d1b7691f0c83319bccf6a9e4b3aeed

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -31,7 +31,7 @@ jobs:
         with:
           dockerfile: Dockerfile
 
-  build:
+  smoke:
     needs: lint
     runs-on: ubuntu-latest
     timeout-minutes: 10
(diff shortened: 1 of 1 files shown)


stranske:
@codex fix the errors below
The failing job (50979949346) in the CI workflow encountered two test failures:

Failure Details:

Both failures are from tests/test_config_models_fallback_loader.py:
test_fallback_loader_activates_when_imports_fail
test_fallback_loader_returns_minimal_payload
Error: ValueError: version must be a string
Root Cause:
These tests expect the payload's "version" field to always be a string, but the fallback loader sometimes returns a non-string value (possibly None or another type).

Solution:

Enforce String Type for Version Field:
In the fallback loader implementation (likely in src/trend_analysis/config/models.py or a related fallback loader module), ensure that "version" is always set to a string—even if the actual value is missing or unavailable.

Example code fix:

Python
# Wherever the fallback payload is constructed
payload = {
    # ... other fields ...
    "version": str(payload.get("version", "")),
}
Or, if you control the fallback loader function:

Python
def fallback_loader(...):
    # ... other logic ...
    version = payload.get("version")
    if not isinstance(version, str):
        payload["version"] = str(version) if version is not None else ""
    # ... rest of function ...
Update Tests if Needed:
If the tests expect a specific string value when imports fail, ensure your fallback loader matches this expectation. For example, set "version": "" or "version": "unknown" if no version is available.

Files to Update:

Check and update the fallback loader logic in [src/trend_analysis/config/models.py](https://github.com/stranske/Trend_Model_Project/blob/e39e463afea146be9122ce4b4c7e6e1ba10fd6b3/src/trend_analysis/config/models.py).
Summary:
Force the "version" field to be a string in the fallback loader payload to resolve the test failures. This will pass the tests and prevent ValueError.

PR #1520, opened 2025-09-22
Docker smoke reliability: buildx cache + timeout - Source Issue #1422

stranske:
## Summary
* Renamed the Docker workflow’s build job to `smoke` to align with the smoke-test scope while retaining the 10-minute timeout guarantee.
* Re-keyed the buildx cache on the runner OS plus the Dockerfile and requirements.lock hashes and added a summary step that logs cache hits or misses to both workflow notices and the job summary.
* Added an explicit `actions/cache/save` step that uploads refreshed buildx layers whenever the cache key misses so subsequent runs can reuse the warmed cache.

## Testing
* `./actionlint -color .github/workflows/docker.yml`

------
https://chatgpt.com/codex/tasks/task_e_68d1baf46dc88331bee1eb7a38888565

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -96,6 +96,13 @@ jobs:
         run: |
           rm -rf /tmp/.buildx-cache
           mv /tmp/.buildx-cache-new /tmp/.buildx-cache
+
+      - name: Save buildx cache
+        if: steps.buildx-cache.outputs.cache-hit != 'true'
+        uses: actions/cache/save@v4
+        with:
+          path: /tmp/.buildx-cache
+          key: ${{ format('buildx-{0}-{1}', runner.os, hashFiles('Dockerfile', 'requirements.lock')) }}
       - name: Run test suite
         run: |
           # Use verbose output only for debug builds, quiet mode for production


stranske:
@codex fix issue below
The failing job 50980849014 in your CI workflow is caused by an unused "type: ignore" comment at line 28 in src/trend_analysis/config/models.py. The mypy error is:

Code
src/trend_analysis/config/models.py:28: error: Unused "type: ignore" comment  [unused-ignore]
Solution:
Remove the # type: ignore[misc] comment from line 28 in models.py, since it is not needed. The code at line 28 is:

Python
def _fallback_validate_trend_config(  # type: ignore[misc]
    data: dict[str, Any], *, base_path: Path
) -> Any:  # pragma: no cover - exercised only in absence
    version = data.get("version")
    if not isinstance(version, str):
        raise ValueError("version must be a string")
    return {
        "version": version,
        "data": data.get("data", {}),
        "portfolio": data.get("portfolio", {}),
        "vol_adjust": data.get("vol_adjust", {}),
    }
Updated code (remove the type ignore comment):

Python
def _fallback_validate_trend_config(
    data: dict[str, Any], *, base_path: Path
) -> Any:  # pragma: no cover - exercised only in absence
    version = data.get("version")
    if not isinstance(version, str):
        raise ValueError("version must be a string")
    return {
        "version": version,
        "data": data.get("data", {}),
        "portfolio": data.get("portfolio", {}),
        "vol_adjust": data.get("vol_adjust", {}),
    }
After removing the unused comment, the mypy check should pass.
See the file in your repository: [src/trend_analysis/config/models.py (ref: 248ebb869dbf12331735901bb4523a44ece985f4)](https://github.com/stranske/Trend_Model_Project/blob/248ebb869dbf12331735901bb4523a44ece985f4/src/trend_analysis/config/models.py).

PR #1618, opened 2025-09-25
ci: add label rules checkout guard

stranske:
## Summary
- add a dedicated guard script to validate the sparse checkout used by the label-agent workflow
- ensure allowlisted paths exist and prevent extra files from slipping through before labels are applied

## Testing
- python -m compileall .github/scripts/label_rules_assert.py


stranske:
@codex fix comments

PR #1648, opened 2025-09-28
test(autofix): cover pipeline tools end-to-end

stranske:
## Summary
- add a synthetic repo fixture that lets us exercise each standalone autofix helper end-to-end
- cover the numpy assert rewriter, expectation refresh, type hygiene ignore injector, mypy return fixer, and cosmetic aggregate normaliser in fresh scenarios
- verifies each helper modifies the expected artefacts so the automation pipeline stays green when new regressions appear

## Testing
- pytest tests/test_autofix_pipeline_tools.py

### DIFF
--- a/.github/workflows/autofix.yml
+++ b/.github/workflows/autofix.yml
@@ -41,6 +41,7 @@ jobs:
       unsafe_change_count: ${{ steps.info.outputs.unsafe_change_count }}
       all_safe: ${{ steps.info.outputs.all_safe }}
       has_opt_in: ${{ steps.info.outputs.has_opt_in }}
+      has_patch_label: ${{ steps.info.outputs.has_patch_label }}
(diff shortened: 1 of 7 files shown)


stranske:
@codex fix the error below
The failing job is caused by a mypy type error:

src/trend_analysis/automation_multifailure.py:14:12: error: Incompatible return value type (got "str", expected "int") [return-value]

The function aggregate_numbers is annotated to return int but actually returns a string. To resolve this, update the return type annotation to str:

Python
def aggregate_numbers(values: Iterable[int]) -> str:
    """Return a pipe-separated string while claiming to return an int."""
    return " | ".join(str(v) for v in values)
See the code on this ref: 25718e2250ec1303a399f15f2077c2e1c52fedc5.

PR #1652, opened 2025-09-29
Add autofix pipeline integration test for live docs

stranske:
## Summary
- add an integration test that mutates live documentation assets and runs the autofix toolchain end-to-end
- seed multiple expected failure modes (typing import, numpy assert, yaml strip, expectation drift, return type) and confirm the scripts repair them
- ensure the automation module and expectation fixtures are restored after the pipeline completes

## Testing
- pytest tests/test_autofix_pipeline_live_docs.py::test_autofix_pipeline_repairs_live_documents

### DIFF
--- a/tests/test_autofix_pipeline_live_docs.py
+++ b/tests/test_autofix_pipeline_live_docs.py
@@ -1,0 +1,279 @@
+from __future__ import annotations
+
+import importlib
+import shutil
+import subprocess
+import sys
+from pathlib import Path
+
+import pytest
+
+import scripts.auto_type_hygiene as auto_type_hygiene
+import scripts.fix_cosmetic_aggregate as fix_cosmetic_aggregate
+import scripts.fix_numpy_asserts as fix_numpy_asserts
+import scripts.mypy_autofix as mypy_autofix
+import scripts.mypy_return_autofix as mypy_return_autofix
+import scripts.update_autofix_expectations as update_autofix_expectations
+
+
+def _run(
+    cmd: list[str],
+    cwd: Path,
+    *,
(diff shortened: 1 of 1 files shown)


stranske:
@codex fix error below
The failing job in your workflow is caused by the test failure in tests/test_autofix_pipeline_live_docs.py, specifically in test_autofix_pipeline_repairs_live_documents. The test asserts that EXPECTED_AUTOFIX_SELECTED_FUNDS is 2 (line 255), but your test setup forcibly sets it to 0 (line 99) during file modification:

Python
if line.startswith("EXPECTED_AUTOFIX_SELECTED_FUNDS"):
    modified_lines.append("EXPECTED_AUTOFIX_SELECTED_FUNDS = 0")
    continue
This causes the final assertion:

Python
assert module.EXPECTED_AUTOFIX_SELECTED_FUNDS == 2
to fail, since the value is 0 instead of the expected 2.

Solution:

Update the line modifying EXPECTED_AUTOFIX_SELECTED_FUNDS in tests/test_autofix_pipeline_live_docs.py (around line 99) to set the value to 2:
Python
if line.startswith("EXPECTED_AUTOFIX_SELECTED_FUNDS"):
    modified_lines.append("EXPECTED_AUTOFIX_SELECTED_FUNDS = 2")
    continue
After making this change, the test should pass and the job will succeed. For full details, see the test file:
[tests/test_autofix_pipeline_live_docs.py (ref: 5dc38d4b94e475f45cef4d182a2813945534a12a)](https://github.com/stranske/Trend_Model_Project/blob/5dc38d4b94e475f45cef4d182a2813945534a12a/tests/test_autofix_pipeline_live_docs.py).

PR #1693, opened 2025-09-29
chore: enforce external joblib resolution on startup

stranske:
## Summary
- Renamed the legacy in-repo `joblib` stub to `joblib_shim.py` and documented its explicit usage so the standard `import joblib` resolves to the real dependency package.
- Added a regression test that confirms `joblib.__file__` points to site/dist-packages rather than the repository tree.
- Added an explicit `__all__` in `joblib_shim.py` so only the shimmed helpers are exported when the module is imported directly.
- Strengthened the joblib safety regression tests to reuse a common location assertion and verify both `__file__` and the importlib spec resolve outside the repository.
- Enforced at interpreter startup that `import joblib` resolves to the external dependency by validating the import spec in `sitecustomize`.

## Testing
- `pytest tests/test_joblib_import.py`

------
https://chatgpt.com/codex/tasks/task_e_68da91e912308331b8fe6debc26aff90

### DIFF
--- a/sitecustomize.py
+++ b/sitecustomize.py
@@ -42,11 +42,8 @@ def _ensure_joblib_external() -> None:
             f"stub at {resolved}."
         )
 
-    if not any(indicator in resolved.parts for indicator in SITE_INDICATORS):
-        raise ImportError(
-            f"joblib resolved to unexpected location {resolved}; expected a "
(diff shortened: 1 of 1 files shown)


stranske:
@codex fix comments

PR #1759, opened 2025-09-30
feat: gate sitecustomize bootstrap behind env flag

stranske:


stranske:
@codex address the causes of the failing test as indicated below
The failing job is caused by assertion errors in the file:

tests/test_sitecustomize_joblib_guard.py
Failure Details
Test Failure 1:

Test: test_maybe_apply_is_noop_without_opt_in
Error:
Python
AssertionError: assert '/home/runner/work/Trend_Model_Project/Trend_Model_Project/src' not in ['/home/runner/work/Trend_Model_Project/Trend_Model_Project', ...]
The test expects that the /src path should not be in sys.path.
Test Failure 2:

Test: test_maybe_apply_inserts_src_when_enabled
Error:
Python
AssertionError: assert '/home/runner...Model_Project' == '/home/runner...l_Project/src'
The test expects /src to be present but finds a different path.
Solution
The problem is with how the source path is set or checked in the tests and/or the implementation. The /src directory is either incorrectly appended or not appended to sys.path, depending on the opt-in configuration.

Code Suggestions
Check the implementation of the function being tested (maybe_apply):

Ensure that when "opt-in" is enabled, /src is correctly appended to sys.path.
When not opted-in, /src should not be present.
Update the Tests for Robustness: In tests/test_sitecustomize_joblib_guard.py, update the assertions to use os.path.abspath and check for equivalence, not substring matches.

Python
import os

def test_maybe_apply_is_noop_without_opt_in():
    src_path = os.path.abspath(os.path.join(os.path.dirname(__file__), "../../src"))
    assert src_path not in map(os.path.abspath, sys.path)

def test_maybe_apply_inserts_src_when_enabled():
    src_path = os.path.abspath(os.path.join(os.path.dirname(__file__), "../../src"))
    assert src_path in map(os.path.abspath, sys.path)
Verify Path Construction:

In your code that modifies sys.path, ensure you use os.path.abspath to build the paths before appending/checking.
Python
import sys
import os

src_dir = os.path.abspath(os.path.join(os.path.dirname(__file__), "../../src"))
if opt_in:
    if src_dir not in map(os.path.abspath, sys.path):
        sys.path.insert(0, src_dir)
Next Steps
Review both the test and the implementation for how paths are constructed and compared.
Update both places to use absolute paths.

PR #1954, opened 2025-10-03
One report builder (HTML/PDF) shared by CLI and app - Source Issue #1683

stranske:
**Summary**
* Introduced a `trend.reporting` module that produces deterministic HTML reports and optional PDF exports for both the CLI and Streamlit workflows.
* Enhanced the `trend report` command with `--output`/`--pdf` options, unified report generation, and shared path resolution while preserving existing artefacts.
* Added Streamlit download buttons wired to the shared generator, broadened unit coverage for report/CLI flows, and declared the `fpdf2` dependency.
* Extended CLI report command tests to cover PDF success output and dependency-failure handling.

**Testing**
* ✅ `pytest tests/test_unified_report.py tests/test_trend_cli.py::test_main_report_supports_output_file_only tests/test_trend_cli.py::test_main_report_uses_requested_directory tests/test_trend_cli.py::test_main_report_writes_pdf_when_requested tests/test_trend_cli.py::test_main_report_pdf_dependency_error tests/test_trend_cli_entrypoints.py::test_main_report_command -q`

------
https://chatgpt.com/codex/tasks/task_e_68df830644108331ba8170ce041ce3db

### DIFF
--- a/tests/test_trend_cli.py
+++ b/tests/test_trend_cli.py
@@ -217,6 +217,81 @@ def test_main_report_supports_output_file_only(monkeypatch, tmp_path: Path) -> N
     assert target.read_text(encoding="utf-8") == "<html>report</html>"
 
 
+def test_main_report_writes_pdf_when_requested(monkeypatch, tmp_path: Path) -> None:
+    dummy_result = RunResult(pd.DataFrame(), {}, 42, {})
+
+    def fake_run_pipeline(*args, **kwargs):  # type: ignore[override]
+        return dummy_result, "pdf001", None
+
+    pdf_bytes = b"%PDF-1.4\n..."
+    recorded: dict[str, object] = {}
+
+    monkeypatch.setattr("trend.cli._ensure_dataframe", lambda _p: pd.DataFrame())
+    monkeypatch.setattr("trend.cli._run_pipeline", fake_run_pipeline)
+    monkeypatch.setattr("trend.cli._print_summary", lambda *args, **kwargs: None)
+    monkeypatch.setattr("trend.cli._write_report_files", lambda *a, **k: None)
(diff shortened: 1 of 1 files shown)


stranske:
@codex fix the type errors described below

Please find a solution for failing job 51867844838. Use the logs, job definition, and any referenced files where the failure occurred. Keep your response focused on the solution and include code suggestions when appropriate.
The failing job encountered two main issues:

Type Error in src/trend_analysis/export/bundle.py:

Error: Returning Any from function declared to return "list[Any]" [no-any-return]
Solution: Update the function at line 188 in src/trend_analysis/export/bundle.py so that its return type matches the expected list[Any]. Avoid returning a value of type Any; instead, ensure the function returns a list with properly typed elements.
Example fix:

Python
# If your function looks like:
def my_function(...) -> list[Any]:
    ...
    return something  # 'something' is currently of type 'Any'

# Change 'something' to ensure it is a list
return list(something)
# Or, type it appropriately so mypy can validate its type.
Review the implementation and either cast the return value to a list, or adjust the logic so the type checker can verify the correct type.

UnicodeEncodeError in PDF Generation Test:

Error: UnicodeEncodeError: 'latin-1' codec can't encode character '\u2022' ...
Solution: The PDF generation code is trying to encode a character not supported by the default font encoding. To resolve this, use a TTF font that supports Unicode or replace problematic characters with supported ones.
Example fix:

If using fpdf, load a TTF font before writing Unicode text:
Python
pdf.add_font('DejaVu', '', 'DejaVuSans.ttf', uni=True)
pdf.set_font('DejaVu', '', 14)
pdf.cell(0, 10, '• Some Unicode text')  # Now this will work
Alternatively, replace • with * or another supported character if using only built-in fonts.

PR #3411, opened 2025-11-09
Codex/keepalive gate reporting

stranske:


stranske:
@codex fix comments

stranske:
@codex I asked you to undo this change. That means that the outcome should be that a successful gate run is NOT required

Restored the keepalive gate default to require successful gate completion before unblocking automation. [.github/scripts/keepalive_gate.jsL417-L456](https://github.com/stranske/Trend_Model_Project/blob/87bd7010041f7538fe759a212d19642f7c1ea0b6/.github/scripts/keepalive_gate.js#L417-L456)

PR #3899, opened 2025-11-30
fix(workflow): STEP 1 - delete agents-pr-meta to clear corrupted registration

stranske:
## Summary
**STEP 1 of 2**: Delete the corrupted workflow file to force GitHub to remove the broken registration.

## Problem
The workflow has corrupted GitHub registration (ID 201068412):
- Shows `.github/workflows/agents-pr-meta.yml` instead of `Agents PR meta manager`
- No triggers on `pull_request` events
- Runs fail with 'workflow file issue'

## This PR
Deletes the workflow file. After merge, a follow-up PR will restore it to create fresh registration.

## After Merge
Immediately merge STEP 2 PR to restore the workflow.


stranske:
Closing - Agents Guard prevents deletion of protected workflow files. Need alternative approach.

PR #4565, opened 2026-01-27
fix: use gpt-4o-mini for compare verifier

stranske:
## Summary
- switch compare-mode GitHub Models default to gpt-4o-mini to avoid 8k request limit

## Testing
- not run (workflow change only)


stranske:
Closing per request; will not change compare defaults to gpt-4o-mini.

PR #5504, opened 2026-06-04
Issue #5436: reject missing policy aliases

stranske:
Closes #5436

## Summary
- remove undocumented missing-policy alias coercions from util missing-policy validation
- keep empty/None defaulting to drop while restricting valid explicit values to drop, ffill, and zero
- add util/io cross-layer regression tests proving undocumented aliases are rejected consistently

## Validation
- PYTHONPATH=src python -m pytest tests/test_util_missing_additional.py -q
- python -m ruff check src/trend_analysis/util/missing.py tests/test_util_missing_additional.py
- git diff --check

## Deliberate-break Gate
- Temporarily restored the bfill/backfill/both -> ffill alias branch; tests/test_util_missing_additional.py::test_missing_policy_layers_reject_undocumented_aliases failed for those aliases, then the temporary alias was removed and the focused suite reran green.

Note: .venv/bin/python is not present in this worktree, so the focused pytest used the available project Python with PYTHONPATH=src.

PR #5506, opened 2026-06-05
Issue #5438: ignore equal numeric model-state types

### LINKED ISSUE #5438: # Why
## Why
`streamlit_app/state.py:317` runs a strict type check before the numeric tolerance branch: `_values_equal` does `if type(left) is not type(right): return False` (line 317) and only reaches the `numbers.Number` / `math.isclose` branch (lines 319-320) afterward. So `_values_equal(10, 10.0, ...)` returns `False` even though the values are numerically equal. Model states round-trip through JSON via `export_model_state` (`state.py:249`, `json.dumps`) and `import_model_state` (`state.py:263`, `json.loads`), where `10` and `10.0` are both valid JSON numbers, so a value written as int can reload as float (or vice-versa). `diff_model_states` (`state.py:335`) then reports a spurious `changed` diff with `type_changed=True`, which surfaces as a `[type changed]` note in `format_model_state_diff` (`state.py:452`). This is a CURRENT break, reproduced live: `_values_equal(10, 10.0, 1e-9)` → `False`, and `diff_model_states({'k':10}, {'k':10.0})` → `[('k','changed',True)]`.

## Scope
Covers reordering `_values_equal` in `streamlit_app/state.py` so the `numbers.Number` / `math.isclose` branch (lines 319-320) is evaluated BEFORE the strict `type(left) is not type(right)` comparison (line 317), so two numerically-equal numbers of different concrete types (int vs float) compare equal. Boundaries: repo-owned (`streamlit_app/` is repo-owned). Only `_values_equal` ordering and a regression test are in scope.

## Non-Goals
- Scaffold-only or partial completion does NOT count as done: moving the numeric branch but leaving the top-level `type(left) is not type(right)` check in `_walk` (`state.py:375`) and the `type_changed=type(left) is not type(right)` flag (`state.py:421`) such that an int↔float still reports `type_changed=True` does NOT satisfy this issue — the phantom diff must disappear for numerically-equal int/float pairs.
- Do NOT change Mapping/Sequence comparison semantics (lines 322-330) beyond what is required to reach them after the numeric reorder.
- Do NOT alter `export_model_state`/`import_model_state` JSON handling, or relax type-change reporting for genuinely different types (e.g. `"10"` string vs `10` int must still be reported).

## Tasks
- [ ] In `streamlit_app/state.py`, reorder `_values_equal` (around lines 316-320) so the `isinstance(left, numbers.Number) and isinstance(right, numbers.Number)` → `math.isclose(...)` branch is checked BEFORE the `type(left) is not type(right): return False` guard, so numerically-equal int/float pairs return `True`.
- [ ] Verify the `changed`-diff path in `diff_model_states._walk` (`state.py:412` calls `_values_equal`; `state.py:421` sets `type_changed=type(left) is not type(right)`) no longer emits a diff for an int-vs-numerically-equal-float change; if it still flags `type_changed` for equal numbers, guard that flag so numerically-equal numbers are not reported as changed at all.
- [ ] Add a test to `tests/app/test_streamlit_state.py` asserting `state._values_equal(10, 10.0, 1e-9) is True` and that `state.diff_model_states({'k': 10}, {'k': 10.0})` returns an empty list (no phantom `changed`/`type_changed` diff). Extend `test_diff_model_states_flags_type_changes_and_float_tolerance` (`tests/app/test_streamlit_state.py:322`) or add a sibling test.
- [ ] Run `.venv/bin/python -m pytest tests/app/test_streamlit_state.py -q` and confirm green.

## Acceptance Criteria
- [ ] New/updated assertion in `tests/app/test_streamlit_state.py`: `state._values_equal(10, 10.0, 1e-9)` is `True` and `state.diff_model_states({'k': 10}, {'k': 10.0})` returns `[]` — passes via `.venv/bin/python -m pytest tests/app/test_streamlit_state.py -q`.
- [ ] A genuinely different-type change is still reported: `state.diff_model_states({'k': 10}, {'k': "10"})` still returns one diff (string vs int is a real change) — asserted in the same test module.
- [ ] Deliberate-break gate: after the fix, re-introduce the early `if type(left) is not type(right): return False` ahead of the numeric branch in `_values_equal`, run the new test, confirm the `_values_equal(10, 10.0)` / empty-diff assertion FAILS, then revert and confirm green.

## Implementation Notes
- File is repo-owned: `streamlit_app/state.py` is under `streamlit_app/`; fix here.
- Confirmed-green reproduction of the CURRENT break (project venv): loading `streamlit_app/state.py` and calling `_values_equal(10, 10.0, 1e-9)` returns `False`; `_values_equal(10.0, 10.0, 1e-9)` returns `True`; `diff_model_states({'k':10}, {'k':10.0})` returns `[('k', 'changed', True)]`.
- `numbers` and `math` are already imported in the module (used at lines 319-320). The fix is purely a branch-ordering change plus the matching `type_changed` guard.
- Use the project venv (`.venv/bin/python`); the system anaconda Python has a NumPy 1.x/2.x mismatch that prevents importing the module's pandas dependency.
- The JSON round-trip rationale is grounded in `export_model_state` (`state.py:249`) / `import_model_state` (`state.py:263`); the existing round-trip test is `test_export_import_round_trip_preserves_types` (`tests/app/test_streamlit_state.py:233`).

stranske:
Closes #5438

## Summary
- compare non-bool numeric state values with numeric tolerance before strict type checks
- keep real type changes, including string-vs-int values, reported as changed
- add regression coverage for int/float JSON round-trip phantom diffs

## Validation
- .venv/bin/python -m pytest tests/app/test_streamlit_state.py -q (unavailable: .venv/bin/python missing in this worktree)
- PYTHONPATH=src python -m pytest tests/app/test_streamlit_state.py -q
- python -m ruff check streamlit_app/state.py tests/app/test_streamlit_state.py
- git diff --check

## Deliberate-break Gate
- Temporarily restored the early type(left) is not type(right) guard before numeric comparison; tests/app/test_streamlit_state.py::test_diff_model_states_ignores_equal_int_float_numbers failed on _values_equal(10, 10.0). Restored the fix and reran focused validation green.

### DIFF
--- a/streamlit_app/state.py
+++ b/streamlit_app/state.py
@@ -313,11 +313,15 @@ def _stringify_value(value: Any) -> str:
         return str(value)
 
 
+def _is_numeric_value(value: Any) -> bool:
+    return isinstance(value, numbers.Number) and not isinstance(value, bool)
+
+
 def _values_equal(left: Any, right: Any, float_tol: float) -> bool:
+    if _is_numeric_value(left) and _is_numeric_value(right):
+        return math.isclose(left, right, rel_tol=float_tol, abs_tol=float_tol)
     if type(left) is not type(right):
         return False
-    if isinstance(left, numbers.Number) and isinstance(right, numbers.Number):
-        return math.isclose(left, right, rel_tol=float_tol, abs_tol=float_tol)
     if isinstance(left, Mapping) and isinstance(right, Mapping):
         if set(left.keys()) != set(right.keys()):
             return False
--- a/tests/app/test_streamlit_state.py
+++ b/tests/app/test_streamlit_state.py
@@ -331,6 +331,18 @@ def test_diff_model_states_flags_type_changes_and_float_tolerance() -> None:
     assert diffs[0].type_changed is True
 
 
+def test_diff_model_states_ignores_equal_int_float_numbers() -> None:
+    assert state._values_equal(10, 10.0, 1e-9) is True
+    assert state.diff_model_states({"k": 10}, {"k": 10.0}) == []
+
+    string_diff = state.diff_model_states({"k": 10}, {"k": "10"})
+
+    assert len(string_diff) == 1
+    assert string_diff[0].path == "k"
+    assert string_diff[0].change_type == "changed"
+    assert string_diff[0].type_changed is True
+
+
 def test_format_model_state_diff_returns_copy_ready_text() -> None:
     diffs = [
         state.ModelStateDiff(
--- a/workloop-state.md
+++ b/workloop-state.md
@@ -1,3 +1,13 @@
+## 2026-06-05T01:06Z - opener (codex): issue #5438 numeric state diff
+
+- Repo/issue: `stranske/Trend_Model_Project` #5438 (`int/float model-state diff phantom change`).
+- Branch/worktree: `codex/issue-5438-numeric-state-diff` in `~/.codex/automations/pd-workloop-resume/worktrees/trend-5438-numeric-state-diff`.
+- Selection: raw opener cap was below 5 after cap/drain sweep. Scoped blockers remained LMS #180, Trend #5343, and Trend #5389/#5440. Trend #5436/#5504 had stale `agent:needs-attention` removed after direct evidence showed the review-path concern was fixed and Gate was green; cap-health then classified #5504 as draining. #5437 is linked to merged PR #5505 and awaiting verifier/source-issue disposition, so #5438 was the oldest unlinked implementation candidate outside blockers.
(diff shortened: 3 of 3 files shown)


stranske:
Closer update: rebased PR #5506 onto current phase-3 after #5504 merged and removed generated workloop-state.md from the effective PR diff in commit d90d8042. Effective files are now limited to streamlit_app/state.py and tests/app/test_streamlit_state.py. Validation: PYTHONPATH=src python -m pytest tests/app/test_streamlit_state.py -q (22 passed, 2 existing protobuf warnings); python -m ruff check streamlit_app/state.py tests/app/test_streamlit_state.py (passed); git diff --check origin/phase-3 (passed). Fresh CI is expected to run asynchronously after the force-with-lease push.

PR #5655, opened 2026-06-23
Complete offline stlite demo: wheel closure + requirements/source bundle [#5643]

stranske:
## Summary

`#5649` vendored the stlite runtime and repointed `demo/wasm/index.html` at local assets (and already uses the correct `new URL(…, window.location.href).href` for `pyodideUrl`), but the offline demo **did not actually boot/run** — so this issue is correctly still open pending verifier PASS. I built + staged the demo, served it from a local static server, and drove it in a real browser (Claude-in-Chrome) while watching the network/server access log. Three gaps remained:

1. **Vendored Pyodide set was too lean** (numpy/pandas + deps only). stlite's micropip bootstrap + streamlit pull a much larger closure. Vendored the **complete Pyodide 0.27.2 closure** (altair, jsonschema stack, pillow, protobuf, fastparquet, cramjam, fsspec, narwhals, jinja2, …).
2. **`presentation_safe` requirements had regressed to `["numpy","pandas"]`**, but the boot path imports `pyyaml` and `pydantic` (config layer) and the engine needs `scipy` + `matplotlib`. Restored these to `REQUIREMENTS` and vendored their closures (`pydantic_core`, `scipy`+`openblas`, the matplotlib stack). **`plotly` is intentionally excluded** — it is not a Pyodide package (pure-PyPI, would force a network fetch) and is only used by the secondary **Visualizations** tab; the default Summary results are table-based and render offline. Vendoring plotly from PyPI for the Visualizations tab is a documented follow-up.
3. **`SOURCE_DIRS` omitted first-party packages the boot path imports** — the repo-root `analysis` package and `src/data`, `src/backtest`. Added them so the demo's own code is fully bundled (`from analysis import Results`, etc.).

## Verification (real browser, network observed)

After the fixes the demo **boots fully offline, renders the Portfolio Simulator**, and **"Run Demo" runs the full analysis → Results**: "Demo results loaded — 10 funds — Sharpe -2.04", the Cumulative Out-of-Sample Performance table (1Y/3Y/LOF CAGR/Volatility/Sharpe/Sortino/MaxDD) and metric cards (Total Return 15.0%, CAGR 3.6%, Ann. Vol 3.1%, Sharpe 1.14).

[...]

PR #5669, opened 2026-06-23
fix(mc): degrade Monte Carlo page gracefully when Plotly is absent (offline demo)

stranske:
## Problem

In the offline stlite/Pyodide demo, navigating to **Monte Carlo** crashes the page with a red traceback:

```
ModuleNotFoundError: No module named 'plotly'
  File ".../streamlit_app/monte_carlo_page.py", line 14, in <module>
    import plotly.graph_objects as go
```

`monte_carlo_page` imports Plotly directly **and** transitively (via `trend_analysis.viz` adapters and the chart export bundle). Plotly is intentionally **excluded** from the lean offline demo bundle (the vendored runtime ships `numpy`/`pandas` only), so importing the page there fails and the whole page dies.

## Fix

Guard the numbered entrypoint [`streamlit_app/pages/5_Monte_Carlo.py`](streamlit_app/pages/5_Monte_Carlo.py) — the **sole** importer of `monte_carlo_page`:

- Probe for Plotly. If unavailable, show a clear "requires Plotly, not in this offline build" message instead of importing the Plotly-dependent page.
- If Plotly **is** present (the full app, or any Plotly-enabled build), the page behaves exactly as before.

Contained entirely to the entrypoint — no changes to `monte_carlo_page`, the shared `viz`/chart modules, or `export_bundle`.

## Tests

Adds `test_monte_carlo_entrypoint_degrades_without_plotly` — executes the real entrypoint file with Plotly stubbed absent and asserts it shows the message instead of raising. Full `tests/streamlit/test_mc_page.py` green (17 passed).

## Verification status

- ✅ **Unit-verified**: the entrypoint executes gracefully (message, no exception) with Plotly absent.
- ⚠️ **Browser end-to-end not verifiable on `phase-3` yet** — see below.

## Separate pre-existing issue found (out of scope here)

While verifying, I found the offline demo on `phase-3` **does not boot at all**: streamlit's Pyodide closure isn't fully vendored. The web-worker access log 404s on `protobuf`, `narwhals`, `fastparquet`, `altair`, `jsonschema`, `pillow`, `cramjam`, … — the `vendor/` dir ships only ~9 of the ~20 required wheels (the initial [#5649](https://github.com/stranske/Trend_Model_Project/pull/5649) vendoring; the closure-completion work isn't merged). Boot fails with `ModuleNotFoundError: protobuf` before any page renders. That's a larger, separate effort (vendor the full closure) and is why this fix is unit-verified rather than browser-verified end-to-end. This guard is still correct and kicks in as soon as the demo boots (or on any plotly-absent build).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Bug Fixes**
  * Monte Carlo page now gracefully handles missing Plotly dependency instead of crashing. Users will see an informational message explaining the requirement when Plotly is unavailable.

### DIFF
--- a/streamlit_app/pages/5_Monte_Carlo.py
+++ b/streamlit_app/pages/5_Monte_Carlo.py
@@ -2,9 +2,35 @@
 
 from __future__ import annotations
 
-from streamlit_app.monte_carlo_page import render
+import streamlit as st
+
 from streamlit_app.theme import apply_ds_theme
 
-
 apply_ds_theme()
-render()
+
+# The Monte Carlo page renders Plotly charts and imports Plotly transitively
+# (monte_carlo_page -> viz adapters / chart export bundle). Plotly is
+# intentionally excluded from the lean offline demo bundle (the stlite/Pyodide
+# vendoring ships numpy/pandas only), so importing the page there raises
+# ModuleNotFoundError and the whole page crashes with a red traceback.
+# Degrade gracefully: when Plotly is unavailable, show a clear message instead
+# of crashing. When Plotly is present (the full app or a Plotly-enabled build)
+# the page behaves exactly as before.
+try:
+    import plotly.graph_objects  # noqa: F401  (availability probe only)
+
+    _PLOTLY_AVAILABLE = True
+except ModuleNotFoundError:
+    _PLOTLY_AVAILABLE = False
+
+if _PLOTLY_AVAILABLE:
+    from streamlit_app.monte_carlo_page import render
+
+    render()
+else:
+    st.title("Monte Carlo Simulation")
+    st.info(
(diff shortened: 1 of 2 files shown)


stranske:
Superseded by #5655 (merged): plotly 6.8.0 is now vendored into the offline demo bundle and listed in requirements, with narwhals bumped to 1.15.1 in the lock — so the Monte Carlo page's Plotly import no longer fails offline and the charts render (verified in-browser). The graceful 'Plotly absent' guard this PR added is therefore no longer needed for the offline demo. Closing; the boot fix that the offline demo actually needs is in #5671 (bundle design-system/ds_streamlit). Branch left on origin for reference.

PR #5824, opened 2026-08-12
fix: keep Results banner Sharpe portfolio-level

### LINKED ISSUE #5810: Results banner reports a per-fund Sharpe as the portfolio headline (observed -2.04 vs Summary 1.14)
## Why

The Results page reports a **per-fund** Sharpe ratio as if it were the portfolio's headline result, and it contradicts the Summary tab on the same screen.

Observed live at tip `b45d1780` (Streamlit 1.58.0, bundled `demo returns.csv`, one click of **Run Demo** on the home page):

- Success banner: **`Demo results loaded — 10 funds — Sharpe -2.04.`**
- Summary metric row, same screen: **`Sharpe Ratio 1.14`** (with Total Return 15.0%, CAGR 3.6%, Ann. Volatility 3.1%)
- Trailing-period table, same screen: `1Y -1.00`, `3Y 1.35`, `LOF 1.14`

**`-2.04` matches no portfolio-level statistic anywhere on the page.** The two figures do not merely differ in precision — they have opposite signs, so the first number an allocator reads after a run says the portfolio was badly negative while the portfolio was in fact positive.

Cause, at `streamlit_app/pages/3_Results.py:114-130`:

```python
def _completed_state_sharpe(result: Any) -> float | None:
    """Return portfolio-level Sharpe for the completed-state banner."""
    ...
    portfolio_sharpe = _stats_value(details.get("out_user_stats"), "sharpe")   # :118  correct source
    if portfolio_sharpe is not None:
        return portfolio_sharpe                                                # :120
    metrics = getattr(result, "metrics", None)                                 # :122  fallback
    if isinstance(metrics, pd.DataFrame) and not metrics.empty:
        for column in ("Sharpe", "sharpe"):
            if column in metrics.columns:
                sharpe = pd.to_numeric(metrics[column], errors="coerce")
                sharpe = sharpe[np.isfinite(sharpe)].dropna()
                if not sharpe.empty:
                    return float(sharpe.iloc[0])                               # :129  FIRST FUND's Sharpe
    return None
```

`result.metrics` is the **per-fund** metrics table, so `.iloc[0]` is one arbitrary manager's Sharpe — not the portfolio's. The function's own docstring promises "portfolio-level Sharpe". The demo policy engine produces a result without `out_user_stats`, so it takes this fallback every time.

No existing test catches this because the fallback only executes when `out_user_stats` is absent, and the banner string is not asserted against the Summary figure.

## Scope

Make the completed-state banner either report the true portfolio Sharpe or report no Sharpe at all. Covers `_completed_state_sharpe` and `_completed_state_line` in `streamlit_app/pages/3_Results.py`.

## Non-Goals

- Do **not** change how portfolio Sharpe is computed anywhere in `src/` — the Summary value (1.14) is correct; only the banner's sourcing is wrong.
- Do **not** add `out_user_stats` to the demo engine's output as the fix for this issue (that may be desirable separately, but the banner must be correct regardless of whether that key exists).
- Do **not** redesign the Results page, its tabs, or the banner's wording beyond what is needed to stop reporting a wrong number.

## Tasks

- [ ] Remove the per-fund fallback at `streamlit_app/pages/3_Results.py:122-129`, or replace it with a genuine portfolio-level computation derived from the same out-of-sample series the Summary tab uses.
- [ ] When no portfolio-level Sharpe is available, omit the `Sharpe …` clause from the banner entirely rather than substituting a per-fund value (`_completed_state_line`, `:100-111`, already handles `sharpe is None`).
- [ ] Audit the sibling helpers in the same file for the same pattern — confirm no other banner/summary field silently falls back to `metrics.iloc[0]`.
- [ ] Add the regression test named in Acceptance Criteria.

## Acceptance Criteria

- [ ] Running the bundled demo (`Run Demo` with the default `Balanced` preset) produces a banner whose Sharpe equals the Summary tab's `Sharpe Ratio` to within display precision, **or** a banner with no Sharpe clause at all.
- [ ] With a `result` whose `details` lacks `out_user_stats`, `_completed_state_sharpe()` never returns a value taken from a single row of `result.metrics`.
- [ ] **Named test gate**: `tests/streamlit/test_results_banner_sharpe.py::test_banner_sharpe_is_portfolio_level_or_absent` builds a result object whose `details` has no `out_user_stats` and whose `metrics` DataFrame has a first-row `Sharpe` of `-2.04` and a portfolio series whose Sharpe is positive; asserts the banner does **not** contain `-2.04`.
- [ ] **Deliberate-break demonstration**: reinstate the `return float(sharpe.iloc[0])` fallback, confirm `test_banner_sharpe_is_portfolio_level_or_absent` **fails**, then revert and confirm it passes. Record both outcomes in the PR description.

## Implementation Notes

- Reproduce by running the app and clicking Run Demo:
  `PYTHONPATH="$PWD/src:$PWD" streamlit run streamlit_app/app.py`
  The default `presentation_safe` profile is sufficient; no data upload or LLM key is needed.
- `_stats_value` (`:133-158`) already filters to finite numerics and is the right helper for a corrected portfolio-level lookup.
- The Summary tab's figure and the `LOF` row of the trailing-period table agreed (both 1.14) in the observed run, so either is a reasonable reference for the test's "portfolio" expectation.

stranske:
Closes #5810

## Summary

- Remove the unsafe per-fund metrics fallback from the completed Results banner.
- Show a Sharpe value only when the result provides an explicit portfolio-level statistic.
- Add a regression test proving a first-row fund Sharpe cannot appear in the banner.

## Validation

- uv run --extra dev python -m pytest -q tests/streamlit/test_results_banner_sharpe.py (1 passed)
- uv run --extra dev ruff check streamlit_app/pages/3_Results.py tests/streamlit/test_results_banner_sharpe.py (passed)

## Deliberate-break evidence

Reinstating the former first-row metrics fallback made the named regression test fail with -2.04 returned; restoring the fix made it pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected the Results-page completion banner to display only the portfolio Sharpe value.
  * Prevented fund-level Sharpe metrics from appearing in the completion-state message.

* **Tests**
  * Added regression coverage to verify the banner excludes fund-level Sharpe values.

### DIFF
--- a/streamlit_app/pages/3_Results.py
+++ b/streamlit_app/pages/3_Results.py
@@ -115,19 +115,7 @@ def _completed_state_sharpe(result: Any) -> float | None:
     """Return portfolio-level Sharpe for the completed-state banner."""
     raw_details = getattr(result, "details", None)
     details = raw_details if isinstance(raw_details, dict) else {}
-    portfolio_sharpe = _stats_value(details.get("out_user_stats"), "sharpe")
-    if portfolio_sharpe is not None:
-        return portfolio_sharpe
-
-    metrics = getattr(result, "metrics", None)
-    if isinstance(metrics, pd.DataFrame) and not metrics.empty:
-        for column in ("Sharpe", "sharpe"):
-            if column in metrics.columns:
-                sharpe = pd.to_numeric(metrics[column], errors="coerce")
-                sharpe = sharpe[np.isfinite(sharpe)].dropna()
-                if not sharpe.empty:
-                    return float(sharpe.iloc[0])
-    return None
+    return _stats_value(details.get("out_user_stats"), "sharpe")
 
 
 def _stats_value(stats: Any, name: str) -> float | None:
--- a/tests/app/test_results_page.py
+++ b/tests/app/test_results_page.py
@@ -707,10 +707,8 @@ def test_results_completed_state_skips_non_finite_fallback_sharpe(
 
     page.render_results_page()
 
-    assert any(
-        "Demo results loaded — 2 funds — Sharpe 1.23." in msg for msg in stub.success_messages
-    )
-    assert not any("Sharpe —" in msg for msg in stub.success_messages)
+    assert any("Demo results loaded — 2 funds." in msg for msg in stub.success_messages)
+    assert not any("Sharpe" in msg for msg in stub.success_messages)
 
 
 def test_demo_run_renders_results_end_to_end(monkeypatch: pytest.MonkeyPatch, results_page) -> None:
--- a/tests/streamlit/test_results_banner_sharpe.py
+++ b/tests/streamlit/test_results_banner_sharpe.py
@@ -12,9 +12,7 @@ def test_banner_sharpe_is_portfolio_level_or_absent(monkeypatch) -> None:
     monkeypatch.setattr(page.st, "session_state", {})
     result = SimpleNamespace(
         details={
-            "out_sample_scaled": pd.DataFrame(
-                {"Fund A": [0.02, 0.01], "Fund B": [0.01, 0.02]}
-            ),
+            "out_sample_scaled": pd.DataFrame({"Fund A": [0.02, 0.01], "Fund B": [0.01, 0.02]}),
             "fund_weights": {"Fund A": 0.5, "Fund B": 0.5},
         },
         metrics=pd.DataFrame({"Sharpe": [-2.04, 1.14]}, index=["Fund A", "Fund B"]),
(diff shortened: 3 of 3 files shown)


stranske:
Runner dispatch state for autofix on PR #5824. Do not edit.

stranske:
Closer recovery: corrected the stale completed-state expectation to omit fund-level Sharpe when `out_user_stats` is absent, and expanded the focused regression to prove the genuine portfolio-level `out_user_stats.Sharpe` path renders `1.14`. Validation on head `8e0c99d7`: `uv run --extra dev python -m pytest -q tests/streamlit/test_results_banner_sharpe.py tests/app/test_results_page.py::test_results_completed_state_skips_non_finite_fallback_sharpe` (2 passed); focused Ruff and diff check passed. The two review findings are addressed; fresh CI and the mandatory post-push review window now apply.

PR #5826, opened 2026-08-12
fix: persist demo profile across page navigation

### LINKED ISSUE #5812: Demo-mode profile: switcher renders only on home page, and ?profile= deep link is lost on first navigation
## Why

Two related defects make the `public_llm_demo` runtime profile effectively unreachable except by one undiscoverable route. Both were observed live at tip `b45d1780`.

**(a) Gated pages instruct the user to use a control that does not exist on those pages.**

- `Model` page in the default `presentation_safe` profile renders: *"Model configuration is unavailable in presentation-safe mode. **Switch to the public LLM demo profile** to enable custom analysis controls."*
- `Data` page renders: *"Custom data upload is disabled in presentation_safe mode. **Switch to public_llm_demo in the sidebar** to upload your own data…"*
- On both pages the sidebar contains **only the seven navigation links** (`app`, `Data`, `Model`, `Results`, `Help`, `Monte Carlo`, `Validation`). There is no `Demo mode` control.

Cause: `render_profile_controls()` is called **only** from `streamlit_app/app.py:41`. No module under `streamlit_app/pages/` calls it, so the switcher exists on the home page and nowhere else.

**(b) The `?profile=` deep link is silently lost on the first in-app navigation.**

- Loading `/Data?profile=public_llm_demo` correctly renders in `public_llm_demo` (upload controls and full fund selection appear).
- Clicking `Model` in the sidebar reverts to `presentation_safe`: *"Model configuration is unavailable in presentation-safe mode."*

Cause: in `streamlit_app/demo_profile.py`, `resolve_profile()` (`:86-94`) has precedence `session > query_param > env`, but the session key `PROFILE_SESSION_KEY` is written **only when the sidebar selectbox value changes** (`render_profile_controls` `:241-242` → `_store_profile_on_module`). A profile arriving via query param is never persisted, and Streamlit drops the query string on page switch, so the next page falls through to `DEFAULT_PROFILE`.

Combined impact: the only working route into `public_llm_demo` is *land on the home page → change the sidebar selectbox → then navigate*. A user who follows the home page's own Quick Start ("Use the **Custom Analysis** section to load your own data") reaches a page that tells them to do something impossible there, and a shared deep link degrades to presentation-safe on the first click. This directly undermines issue #5343 (`repo-review-approved`, `priority:high`), whose stated goal is a browser demo with **switchable** LLM control.

## Scope

Make the active profile persist across in-app navigation regardless of how it was set, and make the profile control reachable from any page that tells the user to change it.

## Non-Goals

- Do **not** change the default profile — `presentation_safe` must remain the default, and an unknown/malformed `?profile=` value must continue to fall back to safe rather than raising (`normalize_profile`, `demo_profile.py:48-61`).
- Do **not** enable LLM or upload surfaces in `presentation_safe`.
- Do **not** bundle provider keys or change the runtime-only key handling.
- Do **not** rewrite the gating copy on every page as a substitute for making the control reachable.

## Tasks

- [ ] Persist a query-param-supplied profile into `PROFILE_SESSION_KEY` on first resolution, so it survives navigation. A natural place is `get_active_profile()` / `_active_profile_from_module()` in `streamlit_app/demo_profile.py` (`:112-150`), or an explicit `set_active_profile()` call at the point the query param is first read.
- [ ] Render the profile switcher on every page that can be gated by it — call `render_profile_controls()` from the page modules (`pages/1_Data.py`, `pages/2_Model.py`, and any other page whose copy references the profile), or hoist it into a shared helper that every page invokes.
- [ ] Ensure the switcher renders exactly once per page (no duplicate `Demo mode` widget, no Streamlit duplicate-key error from the fixed `key="demo_profile_selectbox"`).
- [ ] Re-check each gating message so its instruction matches a control actually present on that page.
- [ ] Add the regression tests named in Acceptance Criteria.

## Acceptance Criteria

- [ ] Loading `/Data?profile=public_llm_demo` and then navigating to `Model` via the sidebar keeps `public_llm_demo` active.
- [ ] Every page whose copy tells the user to switch profile renders a `Demo mode` control.
- [ ] `presentation_safe` remains the default with no query param, env var, or session value set; `?profile=bogus` still resolves to `presentation_safe`.
- [ ] No duplicate-widget-key exception on any page.
- [ ] **Named test gate**: `tests/streamlit/test_demo_profile_persistence.py::test_query_param_profile_persists_across_navigation` resolves a profile from a query param with empty session state, asserts `PROFILE_SESSION_KEY` is then set to `public_llm_demo`, and asserts a subsequent resolution with **no** query param still returns `public_llm_demo`.
- [ ] **Deliberate-break demonstration**: remove the session-persistence write, confirm `test_query_param_profile_persists_across_navigation` **fails**, then revert and confirm it passes. Record both outcomes in the PR description.

## Implementation Notes

- Reproduce with `PYTHONPATH="$PWD/src:$PWD" streamlit run streamlit_app/app.py`, then open `http://127.0.0.1:8501/Data?profile=public_llm_demo` and click `Model` in the sidebar.
- The existing resolution helpers in `demo_profile.py` are intentionally pure and unit-testable without a Streamlit runtime (`resolve_profile`, `normalize_profile`) — the persistence test should target those plus a fake `st_module`, following the pattern the module's own docstring describes.
- `render_profile_controls()` already returns the active profile and is safe to call when Streamlit is unavailable, so calling it from page modules should not require new guards.
- Related and deliberately excluded from this issue: the home page's Quick Start (`app.py:44-53`) unconditionally instructs the user to use a `Custom Analysis` section that `presentation_safe` hides at `:259-264`. Tracked separately.

stranske:
Closes #5812

## Summary
- Persist a valid `?profile=` selection into Streamlit session state before page navigation removes the query string.
- Render the existing Demo mode switcher on the Data and Model pages before their gated-state guidance.
- Cover valid deep-link persistence, malformed-input safety, and gated-page control placement.

## Validation
- `uv run --extra dev pytest -q tests/test_demo_profile.py tests/streamlit/test_data_fund_counter.py` (22 passed)
- `uv run --extra dev ruff check streamlit_app/demo_profile.py streamlit_app/pages/1_Data.py streamlit_app/pages/2_Model.py tests/test_demo_profile.py`
- `git diff --check`

## Deliberate-break evidence
Temporarily disabled the valid-query persistence condition; `tests/test_demo_profile.py::test_query_param_profile_persists_across_navigation` failed with a missing `demo_profile_active` session value. Restoring the condition made the focused suite pass.

### DIFF
--- a/streamlit_app/demo_profile.py
+++ b/streamlit_app/demo_profile.py
@@ -116,23 +116,11 @@ def get_active_profile() -> str:
     only the environment variable and default are consulted.
     """
 
-    session_value: Optional[str] = None
-    query_value: Optional[str] = None
     try:
         import streamlit as st  # noqa: PLC0415 -- lazy, optional dependency
     except ModuleNotFoundError:
-        st = None
-    if st is not None:
-        try:
-            session_value = st.session_state.get(PROFILE_SESSION_KEY)
-        except (AttributeError, TypeError):
-            session_value = None
-        query_value = _query_param_value(st)
-    return resolve_profile(
-        session=session_value,
-        query_param=query_value,
-        env=os.environ.get(PROFILE_ENV_VAR),
-    )
+        return resolve_profile(env=os.environ.get(PROFILE_ENV_VAR))
+    return _active_profile_from_module(st)
 
 
 def _active_profile_from_module(st_module) -> str:
@@ -143,11 +131,21 @@ def _active_profile_from_module(st_module) -> str:
         session_value = st_module.session_state.get(PROFILE_SESSION_KEY)
     except (AttributeError, TypeError):
         session_value = None
-    return resolve_profile(
+    query_value = _query_param_value(st_module)
+    active = resolve_profile(
         session=session_value,
-        query_param=_query_param_value(st_module),
+        query_param=query_value,
         env=os.environ.get(PROFILE_ENV_VAR),
     )
+    # Streamlit drops query parameters when a user changes pages. Persist only
+    # valid deep-link values, leaving malformed input as a safe fallback.
+    if (
+        session_value is None
+        and isinstance(query_value, str)
+        and query_value.strip().lower() in VALID_PROFILES
+    ):
+        _store_profile_on_module(st_module, active)
+    return active
 
 
 def _store_profile_on_module(st_module, profile: str) -> str:
--- a/streamlit_app/demo_profile.py
+++ b/streamlit_app/demo_profile.py
@@ -123,6 +123,23 @@ def get_active_profile() -> str:
     return _active_profile_from_module(st)
 
 
+def initialize_profile(st_module=None) -> str:
+    """Resolve and persist a valid deep-link profile without rendering controls.
+
+    Every Streamlit page must call this at entry so a valid ``?profile=`` is
+    retained even when a visitor lands on a page that does not show the demo
+    mode switcher.  Keeping this separate from :func:`render_profile_controls`
+    lets informational pages preserve the profile without adding duplicate UI.
+    """
+
+    if st_module is None:
+        try:
+            import streamlit as st_module  # noqa: PLC0415
+        except ModuleNotFoundError:
+            return get_active_profile()
+    return _active_profile_from_module(st_module)
+
+
 def _active_profile_from_module(st_module) -> str:
     """Resolve the active profile from an injected Streamlit-like module."""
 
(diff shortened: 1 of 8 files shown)


stranske:
Addressed the profile-persistence review: valid `?profile=` selections are now initialized (without duplicate controls) on Results, Help, Monte Carlo, and Validation; Data and Model retain the visible switcher. The profile-control helper now also falls back to the root Streamlit adapter when a sidebar layout context has no widgets. Validation: focused profile/model tests 103 passed; Ruff and `git diff --check` passed. Full local suite reached 5496 passed, with only two pre-existing optional-LangChain failures because this local `dev` environment lacks `langchain_core`.

PR #5831, opened 2026-08-12
fix: align entrypoint configuration contract

### LINKED ISSUE #5813: Same config, different numbers by entry point: single-period ignores transaction costs and weighting.name; vol_adjust.enabled and harness ddof also split
## Why

The same configuration produces different numbers depending on which entry point runs it. Transaction costs, weighting scheme, and volatility-targeting are each honoured on one path and ignored or stripped on another. For an allocator tool whose selling point is reproducible config-driven runs, this is the highest-consequence class of defect in the repo.

The 2026-06-01 audit found these as single-path defects and they were closed (#5389–#5439). The fixes were applied to the **multi-period** path; the **single-period** path and the strict config model were not brought along. Two independent audit passes (gemini wiring pass, cursor duplication pass) found the cost split separately at tip `b45d1780`.

**1. Transaction costs — single-period path ignores them entirely.**
- Multi-period honours `CostModel` (`src/trend_analysis/multi_period/engine.py:315-350`).
- Single-period deducts only the flat `run.monthly_cost` (`src/trend_analysis/stages/portfolio.py:571-572`, `:803-804`; `src/trend_analysis/pipeline_entrypoints.py:108-117`), so configured `portfolio.transaction_cost_bps` and `portfolio.cost_model.*` have **zero effect** on single-period returns and every metric derived from them. Also reachable via `src/trend_analysis/api.py:488`.

**2. Weighting — `portfolio.weighting.name` is inert on the single-period path.**
- Multi-period resolves both keys (`multi_period/engine.py:178-274`).
- Single-period inspects only `portfolio.weighting_scheme` (`pipeline_entrypoints.py:109`, `stages/portfolio.py:288`) and silently falls back to `"equal"`. `config/defaults.yml:95` ships `portfolio.weighting.name: score_prop_bayes`, which therefore does nothing on that path.

**3. `vol_adjust.enabled` semantics are path-dependent.**
- `pipeline_helpers.py:79` and `:436` **do** read `enabled` from the raw config section.
- The strict config model **pops it**: `config/model.py:551-559`, `_drop_legacy_runtime_flags` removes `enabled` and `window` before validation (the model is `extra="forbid"`, so popping is how the legacy key is tolerated).
- `stages/portfolio.py:472,499` gates volatility scaling on `if target_vol is None:` — not on `enabled`.
- `config/demo.yml:19` sets `enabled: false` with the comment `# keep it off for speed`, so a shipped config expresses an intent whose effect depends on which path consumes it.

**4. The two engines disagree on Sharpe/volatility by construction.**
- `src/trend_analysis/backtesting/harness.py:597` computes both volatility and Sharpe with `active_returns.std(ddof=0)`.
- The canonical metrics use `ddof=1` (`src/trend_analysis/metrics/__init__.py:213`).
- On a 12-period out-of-sample window that is a ~4.4% systematic difference in volatility (`sqrt(12/11)`), so the harness and the main pipeline will report different risk for identical inputs.

None of this is caught by the suite: each path is tested in isolation, and no test asserts that the **same config through different entry points** yields the same numbers.

## Scope

Establish one cost/weighting/vol-adjust contract that every entry point honours identically, and add a cross-entry-point parity test that fails when they diverge. Treat this as **one contract change**, not four separate patches — that is the lesson of the previous fix arc.

## Non-Goals

- Do **not** change the economics of the cost model, the weighting implementations, or the vol-targeting maths. This issue is about **which code reads which key**, not about the formulas.
- Do **not** delete `backtesting/harness.py`. Decide and document whether it is a second engine or a thin delegator; if it must keep `ddof=0` for a stated reason, record that reason in the code and make the difference explicit rather than incidental.
- Do **not** change `config/demo.yml`'s intent (`vol_adjust.enabled: false`) as a way of hiding the inconsistency.
- Do **not** silently start charging costs on paths where a documented decision says they should not apply — if any such decision exists, state it in `Implementation Notes` on the PR and gate accordingly.

## Tasks

- [ ] Route the single-period path through the same cost resolution the multi-period path uses, so `portfolio.transaction_cost_bps` and `portfolio.cost_model.*` take effect (`stages/portfolio.py:571-572,803-804`; `pipeline_entrypoints.py:108-117`).
- [ ] Make the single-period path resolve `portfolio.weighting.name` as well as `portfolio.weighting_scheme`, using the multi-period resolver (`multi_period/engine.py:178-274`) rather than a second copy.
- [ ] Decide the intended semantics of `vol_adjust.enabled` and implement it in one place: either stop popping it at `config/model.py:551-559` and have `stages/portfolio.py` gate on it, or formally deprecate it (warn on use, remove from `config/defaults.yml` and `config/demo.yml`, and drop the `pipeline_helpers.py:79,436` reads). Do not leave both behaviours live.
- [ ] Resolve the `ddof` split: have `harness.py:597` use the canonical metric helpers, or document an explicit, tested reason for `ddof=0`.
- [ ] Add the cross-entry-point parity cases named in Acceptance Criteria to `tests/test_entrypoint_contract_parity.py`.

## Acceptance Criteria

- [ ] A config with a non-zero `portfolio.transaction_cost_bps` produces **lower** net returns than the same config with zero, on the single-period path as well as the multi-period path (direction check).
- [ ] A config with `portfolio.weighting.name: score_prop_bayes` and no `weighting_scheme` produces non-equal weights on the single-period path.
- [ ] `vol_adjust.enabled: false` has one documented, tested meaning that is identical on every path that consumes it.
- [ ] `harness.py` and the canonical metrics report the same Sharpe and volatility for the same return series, or the difference is asserted deliberately by a test that names the reason.
- [ ] **Named test gate**: `tests/test_entrypoint_contract_parity.py::test_same_config_same_numbers_across_entrypoints` runs one fixture config through the single-period and multi-period entry points and asserts the cost-sensitive and weighting-sensitive outputs agree (within a stated tolerance), plus a direction assertion that raising `transaction_cost_bps` lowers net return on **both**.
- [ ] **Deliberate-break demonstration**: revert the single-period cost wiring only, confirm `test_same_config_same_numbers_across_entrypoints` **fails**, then restore and confirm it passes. Record both outcomes in the PR description.

## Implementation Notes

- The fixture must not be homogeneous. The prior audit's own conclusion was that value-level assertions were missing and keys-only checks let wiring bugs survive; use a config that exercises non-default cost bps, a non-equal weighting scheme, and a non-`None` `target_vol`.
- gemini's full wiring pass and cursor's duplication pass are stored at `Code/Audits/Trend_Model_Project/2026-08-11-01-wiring.md` and `…-02-quality-dup.md` and contain the complete line inventories behind the four items above.
- One related claim was **checked and downgraded** during the audit and should not be treated as part of this issue: `data.frequency` is *not* simply "hardcoded to 12" — `infer_periods_per_year()` exists (`util/frequency.py:159`), `regimes.py` threads `periods_per_year` with a frequency-derived default, and `pipeline.py:246` accepts an override. The narrower true statement is that `12` is the metrics-layer default and end-to-end honouring is path-dependent. If you want that closed too, file it separately.

stranske:
Closes #5813

## Summary
- centralize cost and weighting resolution used by both single-period and multi-period entry points
- preserve finite/non-negative cost validation and align harness Sharpe/volatility with the canonical metrics
- add a named cross-entry-point contract test

## Validation
- `PYTHONPATH=src python -m pytest -q tests/test_entrypoint_contract_parity.py tests/test_pipeline_entrypoints.py tests/test_trend_config_model_additional.py tests/test_weighting_resolution.py tests/backtesting/test_harness.py` (66 passed)
- `python -m ruff check src/trend_analysis/config_contract.py src/trend_analysis/pipeline_entrypoints.py src/trend_analysis/multi_period/engine.py tests/test_entrypoint_contract_parity.py`
- `git diff --check`

## Deliberate-break evidence
Temporarily changed the single-period monthly-cost resolver to return `0.0`; `tests/test_entrypoint_contract_parity.py::test_same_config_same_numbers_across_entrypoints` failed with `0.0 == 0.0015`. Restored the shared resolver and the focused suite passed.

## Summary by CodeRabbit

* **Improvements**
  * Standardized portfolio weighting and transaction-cost configuration across single-period and multi-period analysis.
  * Added consistent handling for configuration formats, precedence rules, validation, and legacy fallbacks.
  * Ensured monthly pipeline costs are converted consistently.

* **Bug Fixes**
  * Resolved inconsistencies where equivalent portfolio settings could produce different costs or weighting schemes.
  * Confirmed that nonzero transaction costs appropriately reduce returns.

### DIFF
--- a/src/trend_analysis/config_contract.py
+++ b/src/trend_analysis/config_contract.py
@@ -45,8 +45,8 @@ def resolve_portfolio_weighting_name(
     """Resolve legacy and nested weighting keys with one documented precedence."""
 
     weighting_cfg = section_get(portfolio_cfg, "weighting", {})
-    nested_name = section_get(weighting_cfg, "name")
-    legacy_name = section_get(portfolio_cfg, "weighting_scheme")
+    nested_name = section_get(weighting_cfg, "name", None)
+    legacy_name = section_get(portfolio_cfg, "weighting_scheme", None)
     if legacy_name not in (None, ""):
         resolved_legacy = normalise_weighting_name(legacy_name)
         if resolved_legacy not in _WEIGHTING_SCHEME_PLACEHOLDERS or not nested_name:
@@ -61,9 +61,26 @@ def optional_cost_bps(value: Any, *, field: str) -> float | None:
         parsed = float(value)
     except (TypeError, ValueError) as exc:
         raise CoreConfigError(f"portfolio.{field} must be numeric") from exc
-    if not math.isfinite(parsed) or parsed < 0:
-        raise CoreConfigError(f"portfolio.{field} must be finite and non-negative")
+    if not math.isfinite(parsed):
+        raise CoreConfigError(f"portfolio.{field} must be finite")
+    if parsed < 0:
+        raise CoreConfigError(f"portfolio.{field} cannot be negative")
     return parsed
+
+
+def first_configured_cost(
+    section: Any,
+    primary_key: str,
+    legacy_key: str,
+    *,
+    section_get: SectionGet,
+) -> tuple[Any, str]:
+    """Return the preferred configured cost alias and its diagnostic field name."""
+
+    primary_value = section_get(section, primary_key, None)
+    if primary_value is not None:
+        return primary_value, primary_key
+    return section_get(section, legacy_key, None), legacy_key
 
 
 def resolve_portfolio_cost_bps(
@@ -73,14 +90,26 @@ def resolve_portfolio_cost_bps(
 ) -> tuple[float, float]:
     """Return canonical transaction-cost and slippage inputs in basis points."""
 
-    cost_model = section_get(portfolio_cfg, "cost_model")
+    cost_model = section_get(portfolio_cfg, "cost_model", None)
+    bps_per_trade_value, bps_per_trade_key = first_configured_cost(
+        cost_model,
+        "per_trade_bps",
+        "bps_per_trade",
+        section_get=section_get,
+    )
     bps_per_trade = optional_cost_bps(
-        section_get(cost_model, "per_trade_bps", section_get(cost_model, "bps_per_trade")),
-        field="cost_model.per_trade_bps",
+        bps_per_trade_value,
+        field=f"cost_model.{bps_per_trade_key}",
+    )
+    slippage_bps_value, slippage_bps_key = first_configured_cost(
+        cost_model,
+        "half_spread_bps",
+        "slippage_bps",
+        section_get=section_get,
     )
     slippage_bps = optional_cost_bps(
-        section_get(cost_model, "half_spread_bps", section_get(cost_model, "slippage_bps")),
-        field="cost_model.half_spread_bps",
+        slippage_bps_value,
+        field=f"cost_model.{slippage_bps_key}",
     )
     if bps_per_trade is None:
         bps_per_trade = optional_cost_bps(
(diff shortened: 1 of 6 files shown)


stranske:
Runner dispatch state for autofix on PR #5831. Do not edit.

stranske:
Closer recovery on head 4b36cdb0: optional null cost aliases now fall through to legacy configured fields; empty cost_model preserves run.monthly_cost; explicit zero remains an override; weighting aliases/precedence and invalid finite-cost paths have focused coverage. Validated: 12 focused contract/cost tests passed; three exact prior CI regressions passed; mypy config_contract.py and ruff passed; git diff --check passed. The new exact head is in its seven-minute review window and CI/review checks are running.

Issue #5411, opened 2026-06-03
A31 — Gate developer perf diagnostics out of the production UI (`1_Data.py`)


stranske:
# A31 — Gate developer perf diagnostics out of the production UI (`1_Data.py`)

- **Priority:** P2 · **Area:** streamlit/UX · **Labels:** `ux`, `streamlit` · **Owner:** repo

## Why

`pages/1_Data.py:735-750` renders an always-visible `st.caption` exposing internal perf telemetry to end users (raw ms timings, `init_applied`, `defaults_seeded`, etc.), and a duplicate debug expander already captures the same info at :793-834. **Current** UX leak.

## Scope

Gate the perf caption behind a debug flag (or fold it into the existing expander).

## Non-Goals

- Do NOT remove the debug expander (it is the intended home for this).
- Scaffold-only completion does NOT count: leaving the unconditional caption visible is a failure.

## Tasks

- [ ] In `1_Data.py:735-750`, remove the unconditional perf caption or gate it behind a debug toggle / `st.session_state` debug flag.
- [ ] Add `tests/app/test_data_page_no_perf_caption.py` asserting the perf caption is not rendered when the debug flag is off.

## Acceptance Criteria

- [ ] `python -m pytest tests/app/test_data_page_no_perf_caption.py -q` passes (non-zero collected).
- [ ] **Deliberate-break gate:** temporarily restore the unconditional caption; the new test **must FAIL**. Revert after capture.

## Implementation Notes

- Confirmed: 1_Data.py:735-750 (perf caption), :793-834 (debug expander already present).

stranske:
Sequencing note: PR #5472 merged and now carries `verify:compare`, but this issue remains open until a durable provider-comparison verifier PASS is posted and reviewed by the closer lane.

stranske:
Closer verifier disposition: merged PR #5472 has `verify:compare` and a durable Provider Comparison Report posted 2026-06-03T20:08Z with both providers PASS (openai PASS 96%, anthropic PASS 88%). The report confirms the unconditional Data-page perf caption is now gated behind `show_perf_diagnostics`, the debug surface remains available, and focused regression coverage exists in `tests/app/test_data_page_no_perf_caption.py`. No unresolved review threads or follow-up debt remain, so this issue is complete.

PR #5834, opened 2026-08-12
fix: hide developer surfaces by default

stranske:
Closes #5816

## Summary
- hide Data-page debug state behind the existing diagnostics flag
- keep Quick Start, preset help, validation tooling, and optional LLM diagnostics aligned with the active user experience

## Validation
- `uv run --all-extras pytest tests/app/test_data_page.py::test_debug_surfaces_hidden_without_flag tests/streamlit/test_no_dev_surfaces_by_default.py tests/app/test_preset_vocabulary.py tests/app/test_validation_page_renders.py -q` (6 passed)
- `uv run --all-extras ruff check streamlit_app/app.py streamlit_app/pages/1_Data.py streamlit_app/pages/2_Model.py streamlit_app/pages/4_Help.py streamlit_app/pages/8_Validation.py tests/app/test_data_page.py tests/streamlit/test_no_dev_surfaces_by_default.py`
- `git diff --check`

## Deliberate break
- Replaced the Debug-expander gate with `if True`; the named regression test failed. Restored the gate and reran the passing suite.

## Summary by CodeRabbit

- **User Experience**
  - Quick Start content adapts to the selected demo profile and shows Custom Analysis guidance only when available.
  - Debug and performance diagnostics remain hidden unless explicitly enabled.
  - LLM connection details are available in an optional collapsed panel.

- **Documentation**
  - Clarified the difference between Demo Settings Presets and Model configuration presets.
  - Identified the Validation page as developer-oriented.

- **Tests**
  - Added coverage confirming developer-facing diagnostics and debug surfaces stay hidden by default.

PR #5835, opened 2026-08-13
fix(ui): stop shipping developer surfaces and self-contradictory guidance

### LINKED ISSUE #5816: Developer/QA surfaces exposed to end users, plus self-contradictory Quick Start and duplicate preset vocabularies
## Why

Developer and QA tooling is exposed in the allocator-facing UI, and the app's own guidance contradicts itself in two places. These are individually small, all observed live at tip `b45d1780`, and all cheap to fix.

**1. Ungated debug expander on the Data page.** `streamlit_app/pages/1_Data.py:825` renders `st.expander("Debug: Fund selection state", ...)` unconditionally, dumping internal selection state (`run` counter, `data_loaded_key`, `available_funds_count`, `index_candidates_count`, …) to any user. This is inconsistent with the perf diagnostics in the same file at `:769`, which were correctly gated behind `st.session_state.get("show_perf_diagnostics")` for issue #5411. The gating fix was applied to one diagnostic and missed its sibling.

**2. A QA harness ships in the end-user navigation.** `streamlit_app/pages/8_Validation.py` renders as `Settings Validation` in the sidebar between `Monte Carlo` and nothing, and describes itself as *"This page systematically tests each UI setting to verify it's properly connected to the analysis pipeline."* That is a developer verification tool sitting alongside Data / Model / Results / Help. (Page numbering also skips 6 and 7, suggesting leftover ordering.)

**3. The landing page instructs the user toward a section its own default profile hides.** `streamlit_app/app.py:44-53` renders, unconditionally: *"Use the **Custom Analysis** section to load your own data and configure every parameter."* But `:246-264` only renders that section when `custom_analysis_enabled(profile)` is true, and the default profile is `presentation_safe`, so the same screen instead shows *"🔒 Presentation-safe mode — Custom data upload and LLM analysis are disabled in this mode."* The instruction and the page disagree.

**4. Two preset vocabularies remain user-visible.** `pages/4_Help.py` documents presets as `Baseline / Conservative / Aggressive / Custom`; the home page demo selector (`app.py:73-83`) offers `Balanced`. A user reading Help cannot map its preset table onto the control in front of them. The 2026-06-01 audit flagged this; a static review of this tip concluded it was remediated, but driving the app shows both vocabularies still rendering.

**5. Optional LLM integration presents as a broken required dependency.** On the Model page in `public_llm_demo`, the sidebar leads with `Active provider: OpenAI`, `Active model: gpt-4o-mini`, three failure rows `✗ TS_STREAMLIT_API_KEY`, `✗ TREND_LLM_API_KEY`, `✗ OPENAI_API_KEY`, and the error *"Missing required environment variables for OpenAI."* — placed above the actual configuration controls, for a feature the user may never want.

## Scope

Hide or clearly mark developer/QA surfaces, and make the app's guidance consistent with what the active profile actually renders.

## Non-Goals

- Do **not** delete `pages/8_Validation.py` or the debug expander — gate them. They are useful when diagnosing exactly the kind of wiring bug this repo keeps finding.
- Do **not** remove the LLM settings panel or change how provider keys are handled (runtime-only key handling is deliberate and correct).
- Do **not** rewrite the Help page's content; only reconcile the preset naming.
- Do **not** renumber the `pages/` files as part of this issue unless the routing is unaffected — URL slugs are user-visible.

## Tasks

- [ ] Gate the `Debug: Fund selection state` expander (`1_Data.py:825`) behind the same flag already used for perf diagnostics at `:769` (`show_perf_diagnostics`), or an equivalent debug flag.
- [ ] Gate the `Settings Validation` page behind a debug/developer flag so it does not appear in the default end-user navigation, or retitle it so its developer purpose is unmistakable.
- [ ] Make the home page Quick Start conditional on `custom_analysis_enabled(_active_profile)` so it does not instruct users toward a hidden section (`app.py:44-53` vs `:246-264`).
- [ ] Reconcile preset naming between `pages/4_Help.py` and the demo selector in `app.py:73-83` — one vocabulary, or an explicit note in Help that the demo presets are a separate set.
- [ ] Present the LLM provider/env-var block as an optional-feature affordance (collapsed, or shown only when the user opts into LLM features) rather than a leading error block on the Model page.
- [ ] Add the regression test named in Acceptance Criteria.

## Acceptance Criteria

- [ ] With no debug flag set, the Data page renders no `Debug:` expander.
- [ ] With no debug flag set, `Settings Validation` is absent from the default navigation, or its title marks it as a developer tool.
- [ ] In `presentation_safe`, the home page contains no instruction to use a section that is not rendered.
- [ ] Every preset name shown in the UI appears in Help with the same spelling, or Help states explicitly that demo presets differ from Model presets.
- [ ] The Model page's primary configuration controls are not preceded by an LLM environment-variable error block when the user has not requested LLM features.
- [ ] **Named test gate**: `tests/streamlit/test_no_dev_surfaces_by_default.py::test_debug_surfaces_hidden_without_flag` renders the Data page with a default session state and asserts no rendered text starts with `Debug:`; and `::test_quick_start_matches_active_profile` asserts the home page's rendered text does not reference the `Custom Analysis` section when `custom_analysis_enabled()` is false.
- [ ] **Deliberate-break demonstration**: remove the gate from the debug expander, confirm `test_debug_surfaces_hidden_without_flag` **fails**, then restore and confirm it passes. Record both outcomes in the PR description.

## Implementation Notes

- Reproduce with `PYTHONPATH="$PWD/src:$PWD" streamlit run streamlit_app/app.py`. Items 1, 3, 4 are visible in the default `presentation_safe` profile; items 2 and 5 need the `Demo mode` switcher on the home page set to `Public LLM demo`.
- The perf-diagnostics gate at `1_Data.py:769` is the pattern to copy — its inline comment already states the intent: *"Perf diagnostics are developer-only; gate behind a debug flag so the production UI does not leak raw timings/internal counters (issue #5411)."*
- Full observed evidence, including per-surface aria snapshots and the coverage ledger, is in `Code/Audits/Trend_Model_Project/2026-08-11-UX_REVIEW.md`.
- Related and deliberately excluded: the profile-switcher reachability and `?profile=` persistence defects are tracked in their own issue; the wrong Sharpe banner and wrong fund counter likewise.

stranske:
Closes #5816 — the five UI-honesty items from the 2026-08-11 observed UX review. Every one was found by **driving the live app**, not by reading code.

## What was wrong, and what changed

**1. `Debug: Fund selection state` expander was ungated** (`pages/1_Data.py`)
It dumped internal counters — run counter, `data_loaded_key`, `available_funds_count`, raw perf timings — to every user. #5411 gated the *sibling perf caption* behind `show_perf_diagnostics` but missed this expander, so the same data stayed visible. Now behind the same flag.

**2. `Settings Validation` was indistinguishable from a workflow page** (`pages/8_Validation.py`)
It is a QA harness that checks UI settings reach the pipeline, sitting in the nav beside Data/Model/Results. Legacy `pages/` routing means its file presence creates the nav entry, so the title and intro now state what it is: **"Settings Validation (developer tool)"** plus a developer/QA notice.

**3. Quick Start advertised a section the default profile hides** (`app.py`)
"Use the **Custom Analysis** section to load your own data" rendered unconditionally, while that section is gated on `custom_analysis_enabled()`. In the default `presentation_safe` profile the instruction pointed at a section that is not on the page. Quick Start is now per-profile, and the safe-mode variant points at the **Demo mode** switcher — a control that actually exists.

**4. Two preset vocabularies, unexplained** (`pages/4_Help.py`)
Help documents `Baseline/Conservative/Aggressive/Custom` (Model page) while the home demo selector offers `Balanced`. They are genuinely separate sets, so Help now says so explicitly instead of leaving the user to guess.

**5. An optional LLM integration presented as a broken required dependency** (`pages/2_Model.py`)
The panel led the Model sidebar with `Active provider`, three `✗` rows, and a `st.warning` reading "Missing required environment variables …" — above the actual configuration controls, for a feature the user may never want. It is now a collapsed **"LLM assistant status (optional)"** expander, and an unconfigured provider is a caption. The variable names are still surfaced.

## Test gates

`tests/app/test_no_dev_surfaces_by_default.py`:
- `test_debug_surfaces_hidden_without_flag` — no debug surface without the flag
- `test_debug_surfaces_visible_with_flag` — still reachable for developers
- `test_quick_start_matches_active_profile` — Quick Start copy is profile-gated

**Deliberate-break → revert, all three demonstrated:**

| Break | Result |
|---|---|
| un-gate the Debug expander | `test_debug_surfaces_hidden_without_flag` **FAILS** |
| make Quick Start unconditional | `test_quick_start_matches_active_profile` **FAILS** |
| restore the LLM `st.warning` | `test_llm_status_panel_reports_missing_env_vars_without_warning` **FAILS** |

Each was reverted and the suite returned to green.

## One existing test was retargeted, deliberately

`test_llm_status_panel_warns_once_with_all_missing_env_vars` asserted `st.warning` fires for a missing **optional** dependency — exactly the contract item 5 changes. Rather than delete or weaken it, it now asserts the new contract: the variable names are still shown, "optional" is stated, and `st.warning` is **not** used. Renamed to `test_llm_status_panel_reports_missing_env_vars_without_warning`.

Also note item 2 avoided `st.caption` on the Validation page: that page's existing test stub does not implement `caption`, so the notice went into the `markdown` call it already supports rather than weakening the harness.

## Verification

```
318 passed   (tests/app, tests/streamlit, tests/test_demo_profile)
black --check .                                                   -> 1061 files clean
ruff check --select E4,E7,E9,F --extend-exclude .workflows-lib .   -> All checks passed!
```

Closes #5816

## Summary by CodeRabbit

* **New Features**
  * Quick Start guidance now adapts to the selected demo profile.
  * LLM status details are available in an optional collapsed panel.
  * Help content clarifies that Model and Home presets are separate.

* **Improvements**
  * Developer diagnostics are hidden by default and available when enabled.
  * Validation is clearly identified as a Developer/QA tool.
  * Optional LLM configuration issues no longer appear as warnings.

* **Tests**
  * Added coverage for profile-specific guidance and diagnostic visibility.

### DIFF
--- a/streamlit_app/app.py
+++ b/streamlit_app/app.py
@@ -41,7 +41,13 @@ apply_ds_theme()
 _active_profile = demo_profile.render_profile_controls(st)
 
 st.title("Portfolio Simulator")
-st.markdown("""
+
+# Quick Start must describe only the sections this profile actually renders. The
+# Custom Analysis section below is conditional on ``custom_analysis_enabled``, so
+# advertising it unconditionally sent presentation_safe users looking for a section
+# that is not on the page (issue #5816).
+if demo_profile.custom_analysis_enabled(_active_profile):
+    st.markdown("""
 Welcome! This app analyzes trend-following fund portfolios with volatility adjustment.
 
 **Quick Start:**
@@ -50,6 +56,17 @@ Welcome! This app analyzes trend-following fund portfolios with volatility adjus
 
 The demo uses a specialized policy-based engine optimized for the sample dataset.
 For full control over all parameters, use the Custom Analysis flow.
+    """)
+else:
+    st.markdown("""
+Welcome! This app analyzes trend-following fund portfolios with volatility adjustment.
+
+**Quick Start:**
+- Use the **Demo** section below to run analysis on sample data with preset configurations
+
+This mode runs the bundled synthetic dataset through a deterministic policy-based engine.
+To load your own data and configure every parameter, switch **Demo mode** to
+**Public LLM demo** in the sidebar.
     """)
 
 st.markdown("---")
--- a/streamlit_app/pages/1_Data.py
+++ b/streamlit_app/pages/1_Data.py
@@ -823,49 +823,55 @@ def render_data_page() -> None:
             with summary_cols[2]:
                 st.metric("Fund Columns", final_count)
 
-            # Debug panel to inspect selection state when diagnosing persistence
-            with st.expander("Debug: Fund selection state", expanded=False):
-                st.session_state["_debug_fund_run"] = (
-                    int(st.session_state.get("_debug_fund_run", 0)) + 1
-                )
+            # Debug panel to inspect selection state when diagnosing persistence.
+            # Developer-only: gated behind the same ``show_perf_diagnostics`` flag as the
+            # perf caption above (issue #5411 gated that caption but missed this expander,
+            # so raw internal counters stayed visible to end users -- issue #5816).
+            if st.session_state.get("show_perf_diagnostics"):
+                with st.expander("Debug: Fund selection state", expanded=False):
+                    st.session_state["_debug_fund_run"] = (
+                        int(st.session_state.get("_debug_fund_run", 0)) + 1
+                    )
 
-                snapshot = {
-                    "run": st.session_state.get("_debug_fund_run"),
-                    "data_loaded_key": st.session_state.get("data_loaded_key"),
-                    "available_funds_count": len(available_funds),
-                    "index_candidates_count": len(index_candidates),
-                    "default_selected_funds_count": len(default_selected_funds),
-                    "checkbox_selected_count": len(new_selection_list),
-                    "range_funds_count": len(range_funds),
-                    "defaults_seeded_count": len(defaults),
-                    "perf_ms": {
-                        "derive_funds": round((_t_funds_derived - _t_fund_start) * 1000, 2),
-                        "seed_defaults": round((_t_seed_done - _t_seed_start) * 1000, 2),
-                        "render_checkboxes": round((_t_render_done - _t_render_start) * 1000, 2),
-                        "fund_total": round((_t_render_done - _t_fund_start) * 1000, 2),
-                    },
-                    "selected_fund_columns_count": len(
-                        st.session_state.get("selected_fund_columns") or []
-                    ),
-                    "fund_columns_count": len(st.session_state.get("fund_columns") or []),
-                }
-
-                history = st.session_state.get("_debug_fund_history", [])
-                if not isinstance(history, list):
-                    history = []
-                history.append(snapshot)
-                st.session_state["_debug_fund_history"] = history[-8:]
-
-                st.json(
-                    {
-                        "latest": snapshot,
-                        "history": st.session_state.get("_debug_fund_history", []),
-                        "available_funds": available_funds,
-                        "selected_fund_columns": st.session_state.get("selected_fund_columns"),
-                        "fund_columns": st.session_state.get("fund_columns"),
-                        "checkbox_prefix": include_prefix,
+                    snapshot = {
+                        "run": st.session_state.get("_debug_fund_run"),
+                        "data_loaded_key": st.session_state.get("data_loaded_key"),
+                        "available_funds_count": len(available_funds),
+                        "index_candidates_count": len(index_candidates),
+                        "default_selected_funds_count": len(default_selected_funds),
+                        "checkbox_selected_count": len(new_selection_list),
+                        "range_funds_count": len(range_funds),
+                        "defaults_seeded_count": len(defaults),
+                        "perf_ms": {
+                            "derive_funds": round((_t_funds_derived - _t_fund_start) * 1000, 2),
+                            "seed_defaults": round((_t_seed_done - _t_seed_start) * 1000, 2),
+                            "render_checkboxes": round(
+                                (_t_render_done - _t_render_start) * 1000, 2
+                            ),
+                            "fund_total": round((_t_render_done - _t_fund_start) * 1000, 2),
+                        },
+                        "selected_fund_columns_count": len(
+                            st.session_state.get("selected_fund_columns") or []
+                        ),
+                        "fund_columns_count": len(st.session_state.get("fund_columns") or []),
                     }
-                )
+
+                    history = st.session_state.get("_debug_fund_history", [])
+                    if not isinstance(history, list):
+                        history = []
+                    history.append(snapshot)
+                    st.session_state["_debug_fund_history"] = history[-8:]
+
+                    st.json(
+                        {
+                            "latest": snapshot,
+                            "history": st.session_state.get("_debug_fund_history", []),
+                            "available_funds": available_funds,
+                            "selected_fund_columns": st.session_state.get("selected_fund_columns"),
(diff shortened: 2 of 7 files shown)


stranske:
Closing as **superseded by #5834**, which landed the same five items while this was in progress.

I duplicated work here — I checked for existing work before starting (that is how I confirmed #5810/#5811/#5812 were already done), but #5816 was untouched at that moment and the lane picked it up during implementation. My fault for not re-checking before pushing.

**#5834 is complete, and better than this PR in one respect:** it *renamed* `pages/8_Validation.py` to `streamlit_app/developer_settings_validation.py`, removing the QA harness from the navigation entirely — the stronger of the two options the issue offered. This PR only retitled it.

Item-by-item, #5834 covers all five:

| Item | #5834 |
|---|---|
| Debug expander ungated | gated behind `show_perf_diagnostics` |
| Validation page in end-user nav | **removed from nav** (renamed out of `pages/`) |
| Quick Start advertises a hidden section | gated on `custom_analysis_enabled` |
| Two preset vocabularies | Help now states the Demo preset set is separate |
| LLM optional feature as a required-dependency error | whole panel inside a collapsed "Optional LLM connection status" expander |

On item 5 the only difference is that #5834 keeps `st.warning` for missing vars *inside* the collapsed expander, where this PR downgraded it to a caption. #5834 still satisfies the acceptance criterion — nothing precedes the configuration controls, and the warning is only reached by a user who deliberately opened the LLM panel. That is a preference, not a defect, so it is not worth a follow-up.

**One real gap I will follow up separately:** `tests/streamlit/test_no_dev_surfaces_by_default.py` in #5834 is entirely exact-source-string assertions, including one that pins precise indentation:

```python
assert 'if st.session_state.get("show_perf_diagnostics"):\n                with st.expander("Debug: Fund selection state"' in data_source
```

That breaks on any reformat with no behaviour change, and it never checks that the debug surface actually stays hidden. I will open a small PR adding a behavioural test that drives the Data page and asserts on observable state instead.

stranske:
**Correction to my closing comment above — I mischaracterised #5834's tests and want that on the record.**

I wrote that its test is "entirely exact-source-string assertions" and "never checks that the debug surface actually stays hidden." That is wrong. I only read `tests/streamlit/test_no_dev_surfaces_by_default.py` and missed that #5834 also added `expander_labels` tracking to the `DummyStreamlit` harness plus a genuine behavioural test in `tests/app/test_data_page.py`:

```python
page.render_data_page()
assert not any(label.startswith("Debug:") for label in stub.expander_labels)
```

That drives the real page and asserts on the rendered expander label — a better signal than the `_debug_fund_run` session-key proxy I had used. Apologies for the unfair characterisation.

Two small things do remain, and I will keep the follow-up scoped to just these:

1. **No flag-on counterpart.** Coverage asserts the debug surface is hidden without the flag, but nothing asserts it is still *reachable* with `show_perf_diagnostics` set. As written, deleting the expander outright would also pass — so the test cannot distinguish "correctly gated" from "removed".
2. **Two source assertions pin exact indentation**, e.g.
   ```python
   "never expose selection counters or timing data in the end-user UI.\n"
   '            if st.session_state.get("show_perf_diagnostics"):\n'
   '                with st.expander("Debug: Fund selection state"'
   ```
   Any reformat or re-indent breaks that with no behaviour change. Since the behavioural assertion already covers the intent, these can be relaxed.

PR #5869, opened 2026-08-13
Clarify public APIs and consolidate preset ownership

### LINKED ISSUE #5847: Clarify public APIs and consolidate preset ownership
## Why

`src/health_summarize/__init__.py:393` exports fourteen underscore-prefixed helpers in `__all__`.
Configuration still has confusing parallel public names, and `src/trend_analysis/presets.py:336`
and `signal_presets.py:98` expose separate preset registries used by both CLIs. These are current
ownership/public-surface ambiguities; they should receive explicit dispositions before Phase 6
deletes compatibility exports.

## Scope

Reduce `health_summarize` to demonstrated public symbols, resolve config DTO naming after a caller
inventory, and establish one preset hierarchy/resolver for CLI and UI consumers.

## Non-Goals

Do not merge semantically distinct full-config and signal-only preset payloads without preserving
their type contracts. Do not add compatibility aliases for external users. Scaffold-only completion
does not count: renaming classes or registries while both public sources remain authoritative fails
this issue.

## Tasks

- [ ] Inventory callers of `health_summarize.__all__`, retain only demonstrated public symbols such as `Args`, `build_signature_hash`, and `main`, and move tests to public behavior.
- [ ] Inventory qualified callers of the two `DataSettings` classes and `trend.config_schema.ValidationError`; rename only where it reduces ambiguity, migrating all current callers directly.
- [ ] Define one preset hierarchy/resolver shared by `trend.cli`, `trend_analysis.cli` during migration, and Streamlit; make full-config versus signal-only payloads explicit types rather than parallel authoritative name maps.
- [ ] Retarget preset/config/public-import tests and remove superseded exports/registries.

## Acceptance Criteria

- [ ] `python -m pytest tests/test_health_summarize_module.py tests/test_config_schema.py tests/test_trend_analysis_presets.py tests/test_signal_presets.py tests/app/test_preset_vocabulary.py -q` passes.
- [ ] Every symbol in `health_summarize.__all__` is non-private and covered by a public-behavior test.
- [ ] CLI and Streamlit resolve the same preset name to the same explicit signal/full-config components.
- [ ] Deliberate-break gate: temporarily change the `balanced` signal preset in `signal_presets.py` without changing the canonical resolver; the named CLI/UI parity test must fail, then revert the edit.

## Implementation Notes

This is a public-surface decision issue, not proof that similarly named models or preset types are
runtime duplicates.

stranske:
Closes #5847

## Summary
- make the YAML-backed full preset registry the single owner for CLI and Streamlit resolution
- retain a typed signal-only adapter without a second preset map, and make preset-name matching whitespace tolerant
- reduce health_summarize public exports to non-private entry points with public-contract coverage

## Validation
- python -m pytest tests/test_health_summarize_module.py tests/test_config_schema.py tests/test_trend_analysis_presets.py tests/test_signal_presets.py tests/app/test_preset_vocabulary.py tests/app/test_demo_runner_component.py -q (86 passed)
- python -m ruff check src/trend_analysis/presets.py src/trend_analysis/signal_presets.py streamlit_app/components/demo_runner.py src/health_summarize/__init__.py tests/app/test_preset_vocabulary.py tests/app/test_demo_runner_component.py tests/test_health_summarize_module.py
- python -m black --target-version py312 --check on the touched Python files
- git diff --check
- deliberate break: temporarily changed the signal adapter's balanced specification without changing the canonical resolver; tests/app/test_preset_vocabulary.py::test_cli_and_streamlit_share_preset_resolution failed, then passed after restoration.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -88,6 +88,9 @@ include = [
 # into the package search path when additional roots are added.
 exclude = ["Old*", "retired*", "**/Old*", "**/retired*"]
 
+[tool.setuptools.package-data]
+trend_analysis = ["preset_data/*.yml"]
+
 [project.optional-dependencies]
 app = [
     "streamlit==1.58.0",
--- a/src/health_summarize/__init__.py
+++ b/src/health_summarize/__init__.py
@@ -392,19 +392,6 @@ def main(argv: Sequence[str] | None = None) -> int:
 
 __all__ = [
     "Args",
-    "_branch_row",
-    "_doc_url",
-    "_escape_table",
-    "_extract_contexts",
-    "_format_delta",
-    "_format_require_up_to_date",
-    "_load_json",
-    "_read_bool",
-    "_select_previous_section",
-    "_signature_row",
-    "_snapshot_detail",
-    "_write_json",
-    "_write_summary",
     "build_signature_hash",
     "main",
 ]
--- a/src/trend_analysis/preset_data/aggressive.yml
+++ b/src/trend_analysis/preset_data/aggressive.yml
@@ -1,0 +1,31 @@
+version: "1"
+name: Aggressive
+description: High risk-return profile for performance-oriented users
+lookback_periods: 24
+rebalance_frequency: monthly
+min_track_months: 18
+selection_count: 15
+risk_target: 0.15
+metrics:
+  return_ann: 0.4
+  sharpe_ratio: 0.3
+  max_drawdown: 0.15
+  volatility: 0.15
+signals:
+  window: 42
+  min_periods: 30
+  lag: 1
+  vol_adjust: true
+  vol_target: 0.15
+  zscore: false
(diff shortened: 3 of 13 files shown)


stranske:
Closer recovery pushed in `c26021f8`: canonical preset payloads now deep-copy nested mappings; built wheels include the YAML preset set; malformed/unreadable preset YAML is skipped per file so the Streamlit selector remains usable; and the signal adapter restores its private ordered helpers while deriving values from the canonical registry. Validation: focused preset/demo suite `84 passed`; Ruff, Black, diff check, and wheel-content inspection passed. Full suite was also started locally; its process completed after the focused failures were addressed, but its detached terminal result was not retained by the runner, so CI is the authoritative full-suite confirmation. The exact head is now in the mandatory seven-minute post-push review window.

